PR #1, opened 2023-03-01
Fix directory path to clean modules before installation

khakimosaic:
After restart driver-container or host system, the container fails to install driver because old module exists. 

log:
```
ERROR: The file '/lib/modules/5.19.0-32-generic/kernel/drivers/video/nvidia.ko' already exists as part of this driver installation.
```

So I fix directory path to clean modules before driver installation.


ArangoGutierrez:
Hi @khakimosaic GitHub is a mirror, you are welcome to contribute at https://gitlab.com/nvidia/container-images/driver

ArangoGutierrez:
GitHub is a mirror, please contribute via GitLab
/close

PR #3, opened 2024-01-30
Add Github Actions

ArangoGutierrez:
This patch adds 2 github actions one for pull-request and one for push (post-merge)


.github/workflows/push.yaml
```diff
@@ -0,0 +1,176 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Get kernel version
+        env:
+          BASE_TARGET: jammy
+        run: |
+          trap "docker rm -f base-${BASE_TARGET}" EXIT
+          docker run --pull=always -d --name base-${BASE_TARGET}-${{ matrix.flavor }} registry.gitlab.com/nvidia/container-images/driver/staging/driver:base-${BASE_TARGET}-${{ matrix.flavor }}-${{ matrix.driver }} 
+          # try 3 times every 3 seconds to get the file, if success exit the loop
```
elezar, line 161: Comment is incorrect. There's a 10 second sleep.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,176 @@
...
+    branches:
+      - main
+      - release-*
+
+jobs:
+  image:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        driver: [470.223.02, 525.147.05, 535.129.03]
```
elezar, line 29: Nit: multi-line YAML makes additions and removals easier to review.

ArangoGutierrez: k

ArangoGutierrez:
@cdesiniotis PTAL

.github/workflows/build.yaml
```diff
@@ -0,0 +1,66 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+    branches:
+      - main
+      - release-*
+
+jobs:
```
elezar, line 27: See https://github.com/NVIDIA/libnvidia-container/pull/239 and https://github.com/NVIDIA/nvidia-container-toolkit/pull/332 for a way to selectively run entries in the parallel matrix on PRs. I would rather do that here too than maintain two separate workflows.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,196 @@
...
+      - name: Build image
+        env:
+          IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/driver
+          VERSION: ${COMMIT_SHORT_SHA}
+        run: |
+          make DRIVER_BRANCH=${{ matrix.driver }} KERNEL_FLAVOR=${{ matrix.flavor }} build-base-jammy
+
+  pre-compiled:
+    runs-on: ubuntu-latest
+    needs: base
```
elezar, line 131: Question: Why not have `base` be a step in the precompiled chain instead of a separate job?

elezar (changes requested):

.github/workflows/push.yaml
ArangoGutierrez: I don't see why not, my initial intention was to directly translate the gilab pipeline, and optimize it later down the road. but I like the approach

.github/workflows/build.yaml
ArangoGutierrez: I don't see why not, my initial intention was to directly translate the gilab pipeline, and optimize it later down the road. but I like the approach

.github/workflows/push.yaml
```diff
@@ -0,0 +1,196 @@
...
+          echo "LABEL_IMAGE_SOURCE=https://github.com/${REPO_FULL_NAME}" >> $GITHUB_ENV
+
+          GENERATE_ARTIFACTS="false"
+          if [[ "${{ github.actor }}" == "dependabot[bot]" ]]; then
+            GENERATE_ARTIFACTS="false"
+          elif [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
+            GENERATE_ARTIFACTS="true"
+          elif [[ "${{ github.event_name }}" == "push" ]]; then
+            GENERATE_ARTIFACTS="true"
+          fi
```
cdesiniotis, line 58: Is this if-else block really necessary? I am not very familiar with Github actions, but doesn't the `on` block at the top of this file determine when this job runs? For example, as currently implemented, this job only runs on commits made to `main` and `release-*` branches, right?

ArangoGutierrez: is a work around since we can not push during review process (from a fork) only once is merged (once source is in branch)

.github/workflows/image.yaml
```diff
@@ -0,0 +1,176 @@
...
+              docker cp base-${BASE_TARGET}-${{ matrix.flavor }}:/var/kernel_version.txt kernel_version.txt && break
+              sleep 10
+          done
+      - name: Build image
+        env:
+          IMAGE_NAME: ghcr.io/nvidia/driver
+          VERSION: ${COMMIT_SHORT_SHA}
+          PRECOMPILED: "true"
+          DIST: signed_ubuntu22.04
+          CVE_UPDATES: "curl libc6"
```
elezar, line 173: Nit: Let's drop this.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,176 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build base image and get kernel version
+        env:
+          IMAGE_NAME: ghcr.io/nvidia/driver
+          VERSION: ${COMMIT_SHORT_SHA}
+          BASE_TARGET: jammy
+        run: |
+          make DRIVER_BRANCH=${{ matrix.driver }} KERNEL_FLAVOR=${{ matrix.flavor }} build-base-jammy
```
elezar, line 158: Should `-jammy` be `${BASE_TARGET}`?
```suggestion
          make DRIVER_BRANCH=${{ matrix.driver }} KERNEL_FLAVOR=${{ matrix.flavor }} build-base-${BASE_TARGET}
```

.github/workflows/image.yaml
```diff
@@ -0,0 +1,176 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build base image and get kernel version
+        env:
+          IMAGE_NAME: ghcr.io/nvidia/driver
+          VERSION: ${COMMIT_SHORT_SHA}
+          BASE_TARGET: jammy
+        run: |
```
elezar, line 157: Question: Why are we not using the `precompiled.sh` script here?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,176 @@
...
+          make DRIVER_BRANCH=${{ matrix.driver }} KERNEL_FLAVOR=${{ matrix.flavor }} build-base-jammy
+
+          trap "docker rm -f base-${BASE_TARGET}-${{ matrix.flavor }}" EXIT
+          docker run -d --name base-${BASE_TARGET}-${{ matrix.flavor }} ghcr.io/nvidia/driver:base-${BASE_TARGET}-${{ matrix.flavor }}-${{ matrix.driver }}
```
elezar, line 161: I don't understand why this is needed. The `base/Dockerfile` has a sleep entrypoint. What does running it do?

If we need the file out of it we could run `cp` in a shell or even just `cat /var/kernel_version.txt` if mounts are an issue.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,176 @@
...
+          done
+      - name: Build image
+        env:
+          IMAGE_NAME: ghcr.io/nvidia/driver
+          VERSION: ${COMMIT_SHORT_SHA}
+          PRECOMPILED: "true"
+          DIST: signed_ubuntu22.04
+          CVE_UPDATES: "curl libc6"
+        run: |
+          source kernel_version.txt && \
```
elezar, line 175: Does `kernel_version.txt` need to be "exported" from the previous step in some way?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,176 @@
...
+          make DRIVER_VERSIONS=${DRIVER_VERSIONS} DRIVER_BRANCH=${{ matrix.driver }} build-${DIST}-${DRIVER_VERSION}
```
elezar, line 176: Why don't we reuse the `precompiled.sh` script here?

.github/workflows/image.yaml
ArangoGutierrez: well this was a work around for GitLab pipelines were Docker version and GitLab runner prevented us from doing what you say. so makes sense now that we have a bit more freedom to move to an updated way

.github/workflows/image.yaml
ArangoGutierrez: that script was built to work around GitLab runner capabilities. once we move to Github it can be deprecated

.github/workflows/image.yaml
```diff
@@ -0,0 +1,179 @@
...
+      - main
+      - release-*
+
+jobs:
+  image:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        driver:
+          - 470.223.02
```
cdesiniotis, line 37: Note, 470 is not supported on ubuntu22.04 and rhel9, so we need to skip this version for those distros. See https://gitlab.com/nvidia/container-images/driver/-/blob/main/.common-ci.yml?ref_type=heads#L78 and https://gitlab.com/nvidia/container-images/driver/-/blob/main/.common-ci.yml?ref_type=heads#L91

.github/workflows/image.yaml
```diff
@@ -0,0 +1,179 @@
...
+          - fedora36
+        ispr:
+          - ${{github.event_name == 'pull_request'}}
+        exclude:
+          - ispr: true
+            dist: fedora36
+          - ispr: true
+            dist: rhel8
+          - ispr: true
+            dist: rhel8
```
cdesiniotis, line 55: If I made a change to the `rhel8` folder in a PR, is there anyway to trigger the build for it?

.github/workflows/image.yaml
```diff
@@ -0,0 +1,179 @@
...
+        run: |
+          DRIVER_VERSIONS=${{ matrix.driver }} make build-${{ matrix.dist }}-${{ matrix.driver }}
+
+  pre-compiled:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        driver: 
+          - 535
+          - 550
```
cdesiniotis, line 108: Remove 550 from this matrix. The jobs will continue to fail since no precompiled packages exist for 550 yet.

.github/workflows/image.yaml
elezar: OK. I suppose my question is whether a script makes sense at some point for users / customers to be able to build their own images.

Note a blocker in this PR though.

.github/workflows/image.yaml
elezar: That's a good question. I think we could have excludes based on changed patterns.

.github/workflows/push.yaml
cdesiniotis: Okay. Are you planning on updating this in a follow-up?

.github/workflows/image.yaml
cdesiniotis: @ArangoGutierrez why was this question resolved?

cdesiniotis (approved):
Since the builds are working for 550 and precompiled, I am okay to merge this. We can address any follow-ups / improvements in a follow-up.

.github/workflows/image.yaml
ArangoGutierrez: it should be a follow up

ArangoGutierrez:
> Since the builds are working for 550 and precompiled, I am okay to merge this. We can address any follow-ups / improvements in a follow-up.

Also remember that gitlab pipelines will continue to work in the mean time, so we are double covered for now

elezar (approved):

.github/workflows/image.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+jobs:
+  image:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        driver:
+          - 470.223.02
+          - 535.161.07
+          - 550.54.14
```
tariq1890, line 39: We don't have this released yet. Might want to comment this out before merging

cdesiniotis: It is okay to keep. We have not published it to nvcr.io yet, but it is valid to build 550 images in our GH actions.

tariq1890: I thought we wait till the official release of the first 550 image, but it does make sense to build pre-release images of R550

tariq1890 (approved):

PR #15, opened 2024-03-22
don't hardcode x86_64 when building symlinks against /usr/src/ofa_kernel

tariq1890:


shivamerla (approved):
lgtm, thanks @tariq1890

PR #16, opened 2024-03-25
Remove hardcoded version numbers in Azure Linux precompiled driver container image

henryli001:
- Remove the hardcoded Azure Linux OS version, kernel version, driver version and driver branch. The end users building the image get to supply these values.
- Remove unneeded argument and environment variable definitions


tariq1890:
Thanks @henryli001 !

PR #17, opened 2024-03-26
don't hardcode arch dir when symlinking mellanox ofa_kernel headers

tariq1890:
Same changes as #15 but for `centos7` and `rhel9`. Also includes quoting of paths to avoid word splitting and globs as per shellcheck best practices

See [here](https://github.com/koalaman/shellcheck/wiki/SC2046) for more details on the shellcheck fix


shivamerla (approved):
looks good

PR #21, opened 2024-04-11
[vgpu-manager] update firmware search path so nvidia kernel module ca…

cdesiniotis:
…n find GSP firmware


tariq1890:
Let's do this is in `rhel9` as well please

vgpu-manager/rhel8/nvidia-driver
```diff
@@ -37,6 +37,21 @@ _create_dev_char_directory() {
 	fi
 }
 
+_set_fw_search_path() {
+    local nv_fw_search_path="$RUN_DIR/driver/lib/firmware"
+    local set_fw_path="true"
```
tariq1890, line 42: Are we using this variable?

cdesiniotis: We are not. Good catch. I have removed this.

cdesiniotis:
> Let's do this is in `rhel9` as well please

We currently do not have a `rhel9` directory for the vGPU Manager image.

tariq1890 (approved):

vgpu-manager/ubuntu20.04/nvidia-driver
```diff
@@ -111,6 +111,20 @@ _create_dev_char_directory() {
     fi
 }
 
+_set_fw_search_path() {
+    local nv_fw_search_path="$RUN_DIR/driver/lib/firmware"
+    local fw_path_config_file="/sys/module/firmware_class/parameters/path"
```
shivamerla, line 116: Can we check if the file exists, this may cause an error when this build is used with older gpu-operator versions where sysfs path is not mounted.

cdesiniotis: Note, we already mount the host /sys in the vGPU Manager container, see: https://github.com/NVIDIA/gpu-operator/blob/master/assets/state-vgpu-manager/0500_daemonset.yaml#L61-L62 So no GPU Operator changes are required and thus this change works with existing operator versions.

cdesiniotis: @shivamerla does my comment address your concern? Or would you still like to see a file existence check added here before we merge this.

ArangoGutierrez (changes requested):
Could you re-write the commit message/ PR tittle so it doesn't show as both title and message as is happening right now

cdesiniotis:
> Could you re-write the commit message/ PR tittle so it doesn't show as both title and message as is happening right now

Done.

ping @shivamerla for a re-review.

vgpu-manager/ubuntu20.04/nvidia-driver
shivamerla: yes, if `/sys` was mounted all along then this is fine.

shivamerla (approved):

ArangoGutierrez (approved):

tariq1890:
@cdesiniotis Can this be merged?

PR #22, opened 2024-04-11
remove centos support and bump cuda image base, golang versions

tariq1890:
Looks like cuda version `12.4.1` is no longer available for `centos`, so we remove support for centos


shivamerla:
@tariq1890 can you bump in vgpu folder as well.

tariq1890:
Closing this in favour of #23

PR #23, opened 2024-04-11
remove centos7 support, bump cuda and go versions

tariq1890:
As cuda version `12.4.1` is no longer available for `centos`, we remove support for centos

### DIFF
--- a/.common-ci.yml
+++ b/.common-ci.yml
@@ -112,10 +112,6 @@ trigger-pipeline:
   variables:
     DIST: rhel9
 
-.dist-centos7:
-  variables:
-    DIST: centos7
-
 .dist-fedora36:
   variables:
     DIST: fedora36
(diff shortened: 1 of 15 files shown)


cdesiniotis (approved):

elezar:
As a general comment. Could we keep commits for changes such as these separate. We don't need separate PRs, but reverting or cherrypicking a subset of these changes is simpler if they are separate commits.

.common-ci.yml
```diff
@@ -112,10 +112,6 @@ trigger-pipeline:
   variables:
     DIST: rhel9
 
-.dist-centos7:
```
elezar, line 115: Do we want to add a note to the `centos7` folder in the repo to indicate that it is no longer supported?

vgpu/src/go.mod
```diff
@@ -1,6 +1,6 @@
 module vgpu-util
 
-go 1.20
+go 1.22.2
```
elezar, line 3: Can we revert this instead? This indicate the MINIMUM go version required and does not affect the tooling choice.

PR #29, opened 2024-05-21
vGPU Manager fixes required for vGPU 17.2

cdesiniotis:
See NVBug 4653463 for more details.

### DIFF
--- a/vgpu-manager/rhel8/nvidia-driver
+++ b/vgpu-manager/rhel8/nvidia-driver
@@ -5,6 +5,7 @@ set -xe
 
 DRIVER_VERSION=${DRIVER_VERSION:?"Missing driver version"}
 DRIVER_RESET_RETRIES=10
+DELAY_BEFORE_VF_CREATION=${DELAY_BEFORE_VF_CREATION:-15}
 RUN_DIR=/run/nvidia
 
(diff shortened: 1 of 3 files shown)


vgpu-manager/rhel8/nvidia-driver
```diff
@@ -59,6 +59,11 @@ _load_driver() {
 # Enable virtual functions for all physical GPUs on the node that support SR-IOV.
 # Retry logic is to account for when the driver is busy (i.e. during driver initialization)
 _enable_vfs() {
+    # Wait before attempting to create VFs to ensure the driver has finished initializing.
+    # This is a WAR for a bug in vGPU 17.2 where sriov-manage does not return a non-zero
+    # exit code even though VF creation fails.
+    sleep 15
```
shivamerla, line 65: i see that vGPU team recommended 30s in the nvbug? Should we make this a env variable and set to 15 by default, if that is not sufficient we can let users tweak this as a workaround than having to build the image again. Also, once the fix is available in sriov-manage script, we can default it to 0.

cdesiniotis: Good point. I have made this configurable with an environment variable.

shivamerla (approved):

PR #30, opened 2024-06-05
bump driver versions to 550.90.07, 470.256.02 and 535.183.01

tariq1890:


.github/workflows/image.yaml
```diff
@@ -53,12 +53,12 @@ jobs:
           - ispr: true
             dist: ubuntu20.04
           - ispr: true
-            driver: 470.223.02
+            driver: 470.256.02
```
cdesiniotis, line 56: Question -- do we want to build all active datacenter branches on PRs? Currently the actions that ran only built for R550, so we can't verify that the version bump for R470 and R535 builds okay.

tariq1890: Yeah, I think that's a good idea

tariq1890: I can add those changes to this PR

tariq1890: This is done, please review now

PR #32, opened 2024-06-10
add ngc signing job for auto signing

shivakunv:
@cdesiniotis @elezar PTAL

### DIFF
--- a/.nvidia-ci.yml
+++ b/.nvidia-ci.yml
@@ -419,3 +419,93 @@ release:ngc-rhel9.3:
     - .dist-rhel9
   variables:
     OUT_DIST: "rhel9.3"
+
+# Define the external image signing steps for NGC
+# Download the ngc cli binary for use in the sign steps
+.ngccli-setup:
+  before_script:
+    - apt-get update && apt-get install -y curl unzip jq bash make
+    - export REGCTL_VERSION=v0.4.7
+    - mkdir -p bin
+    - curl -sSLo bin/regctl https://github.com/regclient/regclient/releases/download/${REGCTL_VERSION}/regctl-linux-amd64
+    - chmod a+x bin/regctl
+    - export PATH=$(pwd)/bin:${PATH}
+    - |
+      if [ -z "${NGCCLI_VERSION}" ]; then
+        NGC_VERSION_URL="https://api.ngc.nvidia.com/v2/resources/nvidia/ngc-apps/ngc_cli/versions"
+        # Extract the latest version from the JSON data using jq
+        export NGCCLI_VERSION=$(curl -s $NGC_VERSION_URL | jq -r '.recipe.latestVersionIdStr')
+      fi
+      echo "NGCCLI_VERSION ${NGCCLI_VERSION}"
+    - curl -sSLo ngccli_linux.zip https://api.ngc.nvidia.com/v2/resources/nvidia/ngc-apps/ngc_cli/versions/${NGCCLI_VERSION}/files/ngccli_linux.zip
+    - unzip ngccli_linux.zip
+    - chmod u+x ngc-cli/ngc
+
+# .sign forms the base of the deployment jobs which signs images in the CI registry.
+# This is extended with the image name and version to be deployed.
+.sign:ngc:
+  image: ubuntu:latest
+  timeout: 30 minutes
+  when: always
+  stage: sign
+  rules:
+    - if: $CI_COMMIT_TAG
+      variables:
+        DIST: "${SIGN_JOB_NAME}${VERSION}"
+  variables:
+    OUT_REGISTRY_USER: "${CI_REGISTRY_USER}"
+    OUT_REGISTRY_TOKEN: "${CI_REGISTRY_PASSWORD}"
+    OUT_REGISTRY: "${CI_REGISTRY}"
+    NGC_CLI_API_KEY: "${NGC_REGISTRY_TOKEN}"
+    IMAGE_NAME: "${NGC_REGISTRY_IMAGE}"
+    IMAGE_TAG: "${CI_COMMIT_TAG}"
+  before_script:
+    - !reference [.ngccli-setup, before_script]
+    - 'if [[ "${PRECOMPILED}" != "true" ]]; then echo DRIVER_VERSION: ${DRIVER_VERSION} ; [[ -n "${DRIVER_VERSION}" ]] || exit 1; fi'
+    - regctl registry login "${OUT_REGISTRY}" -u "${OUT_REGISTRY_USER}" -p "${OUT_REGISTRY_TOKEN}"
+    - |
+      if [[ "${PRECOMPILED}" == "true" ]]; then
+        ./scripts/precompiled.sh version;
+        source kernel_version.txt
+        export IMAGE_TAG=${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST##*_}
+      else
+          export IMAGE_TAG=${DRIVER_VERSION}-${DIST##*_}
+      fi
+    # We ensure that the IMAGE_NAME and IMAGE_TAG is set
+    - 'echo Image Name: ${IMAGE_NAME} && [[ -n "${IMAGE_NAME}" ]] || exit 1'
+    - 'echo Image Tag: ${IMAGE_TAG} && [[ -n "${IMAGE_TAG}" ]] || exit 1'
+  script:
+    - 'echo "Signing the image ${IMAGE_NAME}:${IMAGE_TAG}"'
+    - ngc-cli/ngc registry image publish --source ${IMAGE_NAME}:${IMAGE_TAG} ${IMAGE_NAME}:${IMAGE_TAG} --public --discoverable --allow-guest --sign --org nvidia
+
+sign:ngc-precompiled-ubuntu22.04:
+  extends:
+    - .driver-versions-precompiled-ubuntu22.04
+    - .dist-ubuntu22.04
+    - .release-generic
+    - .release:ngc-variables
+    - .sign:ngc
+  variables:
+    DIST: signed_ubuntu22.04
+    BASE_TARGET: jammy
+    PRECOMPILED: "true"
+  needs:
+    - release:ngc-precompiled-ubuntu22.04
+
+sign:ngc-ubuntu-rhel-rhcos:
+  extends:
+    - .sign:ngc
+  parallel:
+    matrix:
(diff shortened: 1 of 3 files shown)


.nvidia-ci.yml
```diff
@@ -419,3 +419,72 @@ release:ngc-rhel9.3:
...
+
+# Define the external image signing steps for NGC
+# Download the ngc cli binary for use in the sign steps
+.ngccli-setup:
+  before_script:
+    - apt-get update && apt-get install -y curl unzip jq bash make
+    - export REGCTL_VERSION=v0.4.7
+    - mkdir -p bin
+    - curl -sSLo bin/regctl https://github.com/regclient/regclient/releases/download/${REGCTL_VERSION}/regctl-linux-amd64
+    - chmod a+x bin/regctl
```
elezar, line 432: Why are we installing `regctl` here too? Looking at the `.sign:ngc` job this seems more complex than the other case and I would have to check to see if there is a way to simplify this somewhat.

.nvidia-ci.yml
```diff
@@ -419,3 +419,72 @@ release:ngc-rhel9.3:
...
+        export IMAGE_TAG=${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST##*_}
+      fi
+    # We ensure that the IMAGE_NAME and IMAGE_TAG is set
+    - 'echo Image Name: ${IMAGE_NAME} && [[ -n "${IMAGE_NAME}" ]] || exit 1'
+    - 'echo Image Tag: ${IMAGE_TAG} && [[ -n "${IMAGE_TAG}" ]] || exit 1'
+  script:
+    - 'echo "Signing the image ${IMAGE_NAME}:${IMAGE_TAG}"'
+    - ngc-cli/ngc registry image publish --source ${IMAGE_NAME}:${IMAGE_TAG} ${IMAGE_NAME}:${IMAGE_TAG} --public --discoverable --allow-guest --sign --org nvidia
+
+sign:ngc-precompiled-ubuntu22.04:
```
elezar, line 478: Are we only signing the precompiled image? What about the other driver cotnainer images?

.nvidia-ci.yml
shivakunv: we are using precompiled.sh to import kernel version and dist tag . 
./scripts/precompiled.sh version;
 source kernel_version.txt
 export IMAGE_TAG=${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST##*_}

precompiled.sh :- 
function sourceVersions(){    --- use regctl 
---- 
if command -v regctl; then    


The KERNEL_VERSION and DIST for pre-compiled images differ from those of regular container images. 
We can simplify the process if we obtain them through other means.

.nvidia-ci.yml
shivakunv: updated the image sign for all available images . 
Please check if I added unnecessary or older images.

.nvidia-ci.yml
```diff
@@ -419,3 +419,241 @@ release:ngc-rhel9.3:
...
+  extends:
+    - .release:ngc
+    - .dist-rhel8
+    - .sign:ngc
+  variables:
+    DIST: "rhel8.9"
+  needs:
+    - release:ngc-rhel8.9
+
+sign:ngc-rhel9.0:
```
cdesiniotis, line 621: Remove all `rhel9.x` jobs since we are not publishing those tags yet.

.nvidia-ci.yml
```diff
@@ -419,3 +419,241 @@ release:ngc-rhel9.3:
...
+
+sign:ngc-ubuntu20.04:
+  extends:
+    - .release
+    - .dist-ubuntu20.04
+    - .sign:ngc
+  needs:
+    - release:ngc-ubuntu20.04
+
+sign:ngc-rhcos4.9:
```
cdesiniotis, line 511: Is it possible to use a matrix so that we have one `sign:ngc-rhcos` job and we specify a matrix of `RHCOS_VERSIONS` (e.g. 4.9, 4.10, ...)? The same suggestion applies for the rhel jobs.

cdesiniotis: This may require us to also parallelize the `release:ngc-rhcos` jobs using similar logic. If required we can improve this in a follow-up.

shivakunv: added : ubuntu 22.04, 20.04 , rhcos and rhel in parallel matrix .
@cdesiniotis check if thel required .
 - SIGN_JOB_NAME: ["rhel" ]
      VERSION: ["8.6", "8.7", "8.8", "8.9","8.10"]
      DRIVER_VERSION: ["470.256.02", "535.183.01", "550.90.07"]

.nvidia-ci.yml
```diff
@@ -419,3 +419,96 @@ release:ngc-rhel9.3:
...
+    - .release:ngc-variables
+    - .sign:ngc
+  variables:
+    DIST: signed_ubuntu22.04
+    BASE_TARGET: jammy
+    PRECOMPILED: "true"
+  needs:
+    - release:ngc-precompiled-ubuntu22.04
+
+sign:ngc-ubuntu-rhel-rhcos:
```
cdesiniotis, line 494: Thanks @shivakunv. This greatly reduces the amount of changes here. The only downside is that we now need to edit the driver versions here as well when bumping the driver versions, like in https://github.com/NVIDIA/gpu-driver-container/commit/65ed7190aa0d9da4b93d7522327f8c29f1f4bfbe. @tariq1890 are you okay if we make this change, or should we revert back to defining a sign job per distro? e.g. `sign:ngc-rhcos4.15`, `sign:ngc-rhcos4.16`, ... ( it is a long list)

tariq1890: This looks good to me!

.nvidia-ci.yml
```diff
@@ -419,3 +419,96 @@ release:ngc-rhel9.3:
...
+  extends:
+    - .sign:ngc
+  parallel:
+    matrix:
+    - SIGN_JOB_NAME: ["ubuntu"]
+      VERSION: ["22.04"]
+      DRIVER_VERSION: ["535.183.06", "550.90.07"]
+    - SIGN_JOB_NAME: ["ubuntu"]
+      VERSION: ["20.04"]
+      DRIVER_VERSION: ["470.256.02", "535.183.01", "550.90.07"]
```
tariq1890, line 504: 535.183.06

.nvidia-ci.yml
```diff
@@ -419,3 +419,96 @@ release:ngc-rhel9.3:
...
+    - SIGN_JOB_NAME: ["rhel"]
+      VERSION: ["8.6", "8.7", "8.8", "8.9","8.10"]
+      DRIVER_VERSION: ["470.256.02", "535.183.01", "550.90.07"]
```
tariq1890, line 507: 535.183.06

.nvidia-ci.yml
```diff
@@ -419,3 +419,96 @@ release:ngc-rhel9.3:
...
+    - SIGN_JOB_NAME: ["rhcos"]
+      VERSION: ["4.9", "4.10", "4.11", "4.12","4.13","4.14","4.15"]
+      DRIVER_VERSION: ["470.256.02", "535.183.01", "550.90.07"]
```
tariq1890, line 510: 535.183.06

.nvidia-ci.yml
shivakunv: corrected @tariq1890

.nvidia-ci.yml
shivakunv: corrected @tariq1890

.nvidia-ci.yml
shivakunv: corrected @tariq1890

shivakunv:
@tariq1890 @cdesiniotis @elezar 

I have tested some new changes  to remove hardcoded value for DRIVER VERSION: 
https://gitlab-master.nvidia.com/dl/container-dev/driver/-/merge_requests/23  

summary : added setup stage to set  the env 

```
stages:
  **- setup**
  - trigger
  - image-build
  - test
  - scan
  - release
  - sign

setup_env:
  stage: setup
  script:
    - apk add --no-cache bash
    - |
      # Ensure compatibility with /bin/sh by using POSIX-compliant syntax
      set -e  # Exit immediately if a command exits with a non-zero status
      IFS=' ' # Set IFS to space

      # Split DRIVER_VERSIONS into individual versions
      i=1  # Initialize a counter
      for version in $DRIVER_VERSIONS; do
        # Construct variable name dynamically
        version_var="DRIVER_VERSION$i"
        echo "$version_var=$version"
        echo "$version_var=$version" >> setup_env.env
        i=$((i + 1))  # Increment counter
      done
    - cat setup_env.env  # Print the contents of setup_env.env for debugging
    - ls -l
  artifacts:
    reports:
      dotenv: setup_env.env
```


```
sign:ngc-ubuntu-rhel-rhcos:
  extends:
    - .sign:ngc
  parallel:
    matrix:
      - SIGN_JOB_NAME: ["ubuntu"]
        VERSION: ["22.04"]
        DRIVER_VERSION: ["$DRIVER_VERSION2", "$DRIVER_VERSION3"]
      - SIGN_JOB_NAME: ["ubuntu"]
        VERSION: ["20.04"]
        DRIVER_VERSION: ["$DRIVER_VERSION1", "$DRIVER_VERSION2", "$DRIVER_VERSION3"]
      - SIGN_JOB_NAME: ["rhel"]
        VERSION: ["8.6", "8.7", "8.8", "8.9", "8.10"]
        DRIVER_VERSION: ["$DRIVER_VERSION1", "$DRIVER_VERSION2", "$DRIVER_VERSION3"]
      - SIGN_JOB_NAME: ["rhcos"]
        VERSION: ["4.9", "4.10", "4.11", "4.12", "4.13", "4.14", "4.15"]
        DRIVER_VERSION: ["$DRIVER_VERSION1", "$DRIVER_VERSION2", "$DRIVER_VERSION3"]
  rules:
    - when: always
      variables:
        DIST: "$SIGN_JOB_NAME${VERSION}"
  needs:
    - setup_env
```


Please suggest what should be done:

_If you are okay with this change, I will merge these changes into the current PR._
or

_I will create a separate Jira ticket for this. I'll remove all hardcoded values and import the DRIVER_VERSION, VERSION, etc., and add them to the setup stage, which can override the current values. In this case, we will have to change only one file or import the pipeline variable from the GitHub variable._

cdesiniotis:
@shivakunv I am fine with the current approach. We can always improve this in a follow-up.

cdesiniotis (approved):

Issue #25, opened 2024-04-25
Add Support For Applying Patches


fifofonix:
For OS versions without official support, e.g. FedoraCoreOS, it may sometimes (^1) be necessary to apply an NVIDIA-forum-recommended patch(es) to overcome compilation issues to deliver a working environment. The current driver container does not expose a means to easily apply patches via the installer's `--apply-patch` switch.

Having a means to`init` the driver-container with an `--apply-patches` switch that searched a specified directory for applicable patches would be a convenience feature for teams running non-supported OS versions. In the situation where the driver container is running on the host one could imagine mounting patches to the container via systemd units etc. However, this method might not extend easily to a `gpu-operator` scheduled driver container (whether there exists an injection point for a) additional command switches, b) additional config files into the k8s/gpu-operator scheduled driver pods needs investigation).

^1 For at least a month it has been known that driver compilation for Fedora40 (which was in beta at the time, but is now released) fails affecting anyone running environments based on  FedoraCoreOS's `next` and now `testing` streams. It is expected that in two weeks unless a formal new driver version is released incorporating the patch, that these failures will migrate to the FedoraCoreOS `stable` stream.

cdesiniotis:
@fifofonix thanks for raising this RFE. I agree that it would be helpful if we exposed an interface for passing additional command line options to the nvidia installer. PRs are always welcome if you are interested in contributing this.

Concerning your second request -- mounting arbitrary config files to the GPU Operator managed driver container -- we would need to evaluate further. cc @shivamerla @tariq1890

fifofonix:
Note I have a work-in-progress pull request for fedora that allows a single patch to be applied to the driver container. With this feature it is possible to get a working fedora40 driver container by applying a specific patch referenced originally on the NVIDIA driver forum.

https://gitlab.com/container-toolkit-fcos/driver/-/issues/11

rajathagasthya:
Since we don't plan to support FCOS as noted in #8, we don't anticipate this will be needed. This has also been stale for a long time and we haven't had any other requests for this enhancement. To keep the issue tracker current, I'm going to close this issue. Please re-open if necessary.

Issue #19, opened 2024-03-27
fedora: typo in kernel-modules-core package name 


unipolar:
https://github.com/NVIDIA/GPU-Driver-Container/blob/main/fedora/nvidia-driver#L132

'-' is missing between 'kernel-modules-core' and '${KERNEL_VERSION}'

cdesiniotis:
cc @fifofonix

PR #36, opened 2024-06-11
Fedora: Add Support For Applying A Single Patch

fifofonix:
# Adding Support For Applying A Single Patch File (#25)

This pull request addresses #25 modifying the container init process to verify whether a patch file has been mounted to the driver container, and if so to apply it as part of driver compilation. This new mode was used successfully until recently for Fedora40 because prior to the availability of the latest driver containers, e.g. 550.90.07.

The pull request also provides a revised systemd unit snippet for FedoraCoreOS in the `README.md` that attempts to look for a kernel-specific image in a registry, but then falls back to a non-kernel-specific image but seeking to mount any available patches that may allow successful driver compilation. 

# Bumping various dependencies to keep Fedora in sync with `main`

Driver, CUDA and Go versions.

# FedoraCoreOS GitLab CICD Changes

Various minor changes to the example GitLab CICD scripting.

# Addressing Typo As Described in #19

### DIFF
--- a/ci/fedora/.gitlab-ci-fcos.yml
+++ b/ci/fedora/.gitlab-ci-fcos.yml
@@ -35,9 +35,9 @@
 # the commit sha).  This release stage is purely to test out the CICD code that
 # would for the 'fedora' branch publish to a remote repository.
 #
-# Branches == "fedora" and tags == .*fedora$
+# Tags == .*fedora$
 #
-# The protected branch 'fedora' will cause container image builds on all three
+# Matching pipelines will cause container image builds on all three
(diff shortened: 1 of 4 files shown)


tariq1890:
Hi @fifofonix thank you for your contribution. After some internal discussions, we have decided to remove the directories of the driver image scripts for the OS versions/distros that we don't actively support (see this [PR](https://github.com/NVIDIA/gpu-driver-container/pull/739)). The gpu-driver-container repo will now only contain the driver image scripts that are actively maintained by us

PR #45, opened 2024-07-08
Switch to using nvcr.io for CUDA base images

cdesiniotis:


tariq1890 (approved):
Thanks @cdesiniotis !

PR #46, opened 2024-07-08
Add dependabot rule for updating CUDA base images

cdesiniotis:


temp/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM nvcr.io/nvidia/cuda:12.5.0-base-ubuntu20.04
```
tariq1890, line 1: Why do we need this?

cdesiniotis: I pushed that by mistake. Thanks for catching. I have removed it now.

tariq1890 (approved):

PR #65, opened 2024-07-17
new branch with linux header changes

JunAr7112:
new branch with linux header changes

### DIFF
--- a/ubuntu22.04/nvidia-driver
+++ b/ubuntu22.04/nvidia-driver
@@ -63,39 +63,40 @@ _resolve_kernel_version() {
 }
 
 # Install the kernel modules header/builtin/order files and generate the kernel version string.
-_install_prerequisites() (
+_install_prerequisites() { 
     local tmp_dir=$(mktemp -d)
 
     trap "rm -rf ${tmp_dir}" EXIT
     cd ${tmp_dir}
-
-    rm -rf /lib/modules/${KERNEL_VERSION}
-    mkdir -p /lib/modules/${KERNEL_VERSION}/proc
-
-    echo "Installing Linux kernel headers..."
-    apt-get -qq install --no-install-recommends linux-headers-${KERNEL_VERSION} > /dev/null
-
-    echo "Installing Linux kernel module files..."
-    apt-get -qq download linux-image-${KERNEL_VERSION} && dpkg -x linux-image*.deb .
-    { apt-get -qq download linux-modules-${KERNEL_VERSION} && dpkg -x linux-modules*.deb . || true; } 2> /dev/null
-    mv lib/modules/${KERNEL_VERSION}/modules.* /lib/modules/${KERNEL_VERSION}
-    mv lib/modules/${KERNEL_VERSION}/kernel /lib/modules/${KERNEL_VERSION}
-    depmod ${KERNEL_VERSION}
+    
+    if [ "${PERSIST_DRIVER}" = false ]; then
+        rm -rf /lib/modules/${KERNEL_VERSION}
+    fi
+    
+    if [ ! -d "/usr/src/linux-headers-$(uname -r)/" ]; then
+        echo "Installing Linux kernel headers..."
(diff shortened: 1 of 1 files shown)


cdesiniotis (commented):
Thanks @JunAr7112 for the progress on this. I left some comments. I mainly have some questions around how to accurately determine whether we need to install these packages or not. Ideally we would have a robust way to determine this and avoid using / introducing new flags / environment variables for this feature.

ubuntu22.04/nvidia-driver
```diff
@@ -63,39 +63,40 @@ _resolve_kernel_version() {
...
 # Install the kernel modules header/builtin/order files and generate the kernel version string.
-_install_prerequisites() (
+_install_prerequisites() { 
     local tmp_dir=$(mktemp -d)
 
     trap "rm -rf ${tmp_dir}" EXIT
     cd ${tmp_dir}
+    
+    if [ "${PERSIST_DRIVER}" = false ]; then
+        rm -rf /lib/modules/${KERNEL_VERSION}
```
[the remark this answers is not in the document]
cdesiniotis: On container startup, if we do not mount `/lib/modules` or `/usr/src` explicitly from the host, both of these directories will be empty inside of the container. As an example:
```
$ kubectl exec -n gpu-operator -it nvidia-driver-daemonset-7qwn5 -- sh
# ls -ltr /lib/modules/
ls: cannot access '/lib/modules/': No such file or directory
# ls -ltr /usr/src/
total 0
```
Based on this observation, what are your thoughts on my below proposal?
1) Remove the following command from our script as it appears to not be needed: `rm -rf /lib/modules/${KERNEL_VERSION}`
2) Only install the linux-modules package if `/lib/modules/${KERNEL_VERSION}` directory does not exist.

ubuntu22.04/nvidia-driver
```diff
@@ -63,39 +63,40 @@ _resolve_kernel_version() {
...
-    mv lib/modules/${KERNEL_VERSION}/kernel /lib/modules/${KERNEL_VERSION}
-    depmod ${KERNEL_VERSION}
-
-    echo "Generating Linux kernel version string..."
+    mv version /lib/modules/${KERNEL_VERSION}/proc
 
-    ls -1 boot/vmlinuz-* | sed  's/\/boot\/vmlinuz-//g' - > version
-    if [ -z "$(<version)" ]; then
-        echo "Could not locate Linux kernel version string" >&2
-        return 1
```
cdesiniotis, line 90: @shivamerla after reviewing this again, do we even really require to install the `linux-image-$KERNEL_VERSION` package? It seems like we only use it to construct the kernel-version string. However, the kernel version string is already assumed to be set correctly in the KERNEL_VERSION environment variable before we ever reach this point in the script.

Based on my understanding, I think we can remove this code block entirely in all cases and never install the `linux-image-$KERNEL_VERSION` package.

ubuntu22.04/nvidia-driver
```diff
@@ -63,39 +63,40 @@ _resolve_kernel_version() {
...
+    if [ "${PERSIST_DRIVER}" = false ]; then
+        mv lib/modules/${KERNEL_VERSION}/modules.* /lib/modules/${KERNEL_VERSION}
+	mv lib/modules/${KERNEL_VERSION}/kernel /lib/modules/${KERNEL_VERSION}
+	depmod ${KERNEL_VERSION}
```
cdesiniotis, line 92: Based on my comment https://github.com/NVIDIA/gpu-driver-container/pull/65/files#r1681872260 I would recommend the following change:
```suggestion
    if [ ! -d "/lib/modules/${KERNEL_VERSION}" ]; then
        { apt-get -qq download linux-modules-${KERNEL_VERSION} && dpkg -x linux-modules*.deb . || true; } 2> /dev/null
        mv lib/modules/${KERNEL_VERSION}/modules.* /lib/modules/${KERNEL_VERSION}
	mv lib/modules/${KERNEL_VERSION}/kernel /lib/modules/${KERNEL_VERSION}
	depmod ${KERNEL_VERSION}
```

PR #88, opened 2024-08-16
end-to-end  gpu driver testing enhancement

shivakunv:
TODO--

*1. *
> + matrix:
+ driver:
+ - 535.183.06
+ - 550.90.07
An idea for a potential follow-up – instead of defining a matrix and spinning up one AWS instance per driver version, can we instead pass all driver versions as input to the test script and test all of them in sequence? For example, first install 535.183.06, and then upgrade to 550.90.07? This would also allow us to source the driver versions from the versions.mk file, instead of having to redefine and maintain that list here.

2
+name: CI
Let's rename this to End-to-end tests

3
+# Install the operator with usePrecompiled mode set to true
Remove this comment as it is not accurate.

4
+ AWS_SESSION_TOKEN: ${{ secrets.AWS_SESSION_TOKEN }}
I don't believe AWS_SESSION_TOKEN is needed with current holodeck implementation. Let's remove.

5
We can simply wait for the nvidia-driver pod to be ready
kubectl wait -n ${TEST_NAMESPACE} --for=condition=Ready pod -l app=nvidia-driver-daemonset --timeout 10m
If successful, then wait for the validator pod to be ready (this means that the rest of the pods are healthy):
kubectl wait -n ${TEST_NAMESPACE} --for=condition=Ready pod -l app=nvidia-operator-validator --timeout 2m
If either of these commands fails, capture the state of all pods in the operator namespace, by running kubectl get pods -n ${TEST_NAMESPACE} , and also capture some logs so we can debug.
This will reduce the amount of logs emitted during the test. Right now, we print out all pods every 5 seconds so it is very unreadable

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: CI
+name: End-to-end tests
 
 on:
   workflow_run:
@@ -25,11 +25,6 @@ on:
 jobs:
   e2e-tests-nvidiadriver:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        driver:
-          - 535.183.06
-          - 550.90.07
 
     steps:
     - name: Check out code
@@ -41,7 +36,6 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SSH_KEY: ${{ secrets.AWS_SSH_KEY }}
-        AWS_SESSION_TOKEN: ${{ secrets.AWS_SESSION_TOKEN }}
       with:
         aws_access_key_id: ${{ secrets.AWS_ACCESS_KEY_ID }}
         aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
@@ -59,6 +53,8 @@ jobs:
         echo "instance_hostname=ubuntu@${{ steps.get_public_dns_name.outputs.result }}" >> $GITHUB_ENV
         echo "private_key=${{ github.workspace }}/key.pem" >> $GITHUB_ENV
         echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
+        DRIVER_VERSIONS=$(grep '^DRIVER_VERSIONS ?=' versions.mk | awk -F' ?= ' '{print $2}')
+        echo "DRIVER_VERSIONS=$DRIVER_VERSIONS" >> $GITHUB_ENV
         
     - name: Validate gpu driver
       env:
@@ -66,4 +62,23 @@ jobs:
       run: |
         sudo chmod 644 ${{ github.workspace }}/.cache/key
         echo "${{ secrets.AWS_SSH_KEY }}" > ${private_key} && chmod 400 ${private_key}
-        ./tests/ci-run-e2e.sh ${TEST_CASE} ${COMMIT_SHORT_SHA}-${{ matrix.driver }}
+        rc=0
+        for driver_version in ${DRIVER_VERSIONS}; do
+          echo "Running e2e for DRIVER_VERSION=$driver_version"
+          ./tests/ci-run-e2e.sh ${TEST_CASE} ${COMMIT_SHORT_SHA}-${driver_version} || status=$?
+          if [ $status -ne 0 ]; then
+            echo "e2e validation failed for driver version $driver_version with status $status"
+            rc=$status
+          fi
+        done
+        source ./tests/scripts/.definitions.sh
+        ./tests/scripts/pull.sh ${LOG_DIR} logs
+        exit $rc
+
+    - name: Archive test logs
+      if: ${{ failure() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: nvidiadriver-e2e-test-logs
+        path: ./logs/
+        retention-days: 15  
--- a/tests/scripts/checks.sh
+++ b/tests/scripts/checks.sh
@@ -2,35 +2,20 @@
 
 check_pod_ready() {
 	local pod_label=$1
-	local current_time=0
-	while :; do
-		echo "Checking $pod_label pod"
-		kubectl get pods -lapp=$pod_label -n ${TEST_NAMESPACE}
+	local pod_status_time_out=$2
+	
+	echo "Checking $pod_label pod"
+	
+	kubectl get pods -lapp=$pod_label -n ${TEST_NAMESPACE}
 
-		echo "Checking $pod_label pod readiness"
-		is_pod_ready=$(kubectl get pods -lapp=$pod_label -n ${TEST_NAMESPACE} -ojsonpath='{range .items[*]}{.status.conditions[?(@.type=="Ready")].status}{"\n"}{end}' 2>/dev/null || echo "terminated")
+	echo "Checking $pod_label pod readiness"
 
-		if [ "${is_pod_ready}" = "True" ]; then
-			# Check if the pod is not in terminating state
-			is_pod_terminating=$(kubectl get pods -lapp=$pod_label -n ${TEST_NAMESPACE} -o jsonpath='{.items[0].metadata.deletionGracePeriodSeconds}' 2>/dev/null || echo "terminated")
-			if [ "${is_pod_terminating}" != "" ]; then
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -22,14 +22,14 @@ on:
     branches:
       - main
 
+  # SHIVA
```
shivakunv, line 25: will remove for testing purpose only

.github/workflows/ci.yaml
```diff
@@ -59,11 +58,34 @@ jobs:
...
         TEST_CASE: "./tests/cases/nvidia-driver.sh"
       run: |
         sudo chmod 644 ${{ github.workspace }}/.cache/key
         echo "${{ secrets.AWS_SSH_KEY }}" > ${private_key} && chmod 400 ${private_key}
-        ./tests/ci-run-e2e.sh ${TEST_CASE} ${COMMIT_SHORT_SHA}-${{ matrix.driver }}
+        rc=0
+        for driver_version in ${{ env.DRIVER_VERSIONS }}; do
+          echo "Running e2e for DRIVER_VERSION=$driver_version"
+          # ./tests/ci-run-e2e.sh ${TEST_CASE} ${COMMIT_SHORT_SHA}-${driver_version}
+          # SHIVA
```
shivakunv, line 73: will remove , for testing purpose

.github/workflows/image.yaml
```diff
@@ -93,86 +93,89 @@ jobs:
           IMAGE_NAME: ghcr.io/nvidia/driver
           VERSION: ${COMMIT_SHORT_SHA}
         run: |
-          DRIVER_VERSIONS=${{ matrix.driver }} make build-${{ matrix.dist }}-${{ matrix.driver }}
+          # SHIVA
+          # DRIVER_VERSIONS=${{ matrix.driver }} make build-${{ matrix.dist }}-${{ matrix.driver }}
```
shivakunv, line 97: for testing purpose

tests/scripts/verify-operator.sh
```diff
@@ -11,9 +11,39 @@ source ${SCRIPT_DIR}/.definitions.sh
...
+# collect log in case of failure
+local start_time=$(date +%s)
+local log_dir=${LOG_DIR}
+while :; do
+    current_time=$(date +%s)
+    elapsed_time=$((current_time - start_time))
+    
+    # Check if total elapsed time is greater than exit
+    if [ $elapsed_time -gt $MAX_POD_STATUS_CHECK_TOTAL_TIME ]; then
+        echo "Total wait time exceeded ${MAX_POD_STATUS_CHECK_TOTAL_TIME} seconds. Exiting..."
```
shivakunv, line 25: exit when 50 minutes completed

tests/scripts/checks.sh
```diff
@@ -2,35 +2,32 @@
...
-		if [[ "${current_time}" -gt $((60 * 45)) ]]; then
-			echo "timeout reached"
-			exit 1;
-		fi
+	return 1
+}
 
-		# Echo useful information on stdout
-		kubectl get pods -n ${TEST_NAMESPACE}
+cleanup_pod(){
```
cdesiniotis, line 23: Instead of implementing this function, create a new script file named `uninstall-operator.sh` that uninstalls the operator. Uninstalling the chart will remove all of the resources in the namespace, so you just need to run:

```
helm uninstall $release_name --namespace $namespace
```

tests/scripts/remote.sh
```diff
@@ -4,4 +4,5 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source ${SCRIPT_DIR}/.definitions.sh
 source ${SCRIPT_DIR}/.local.sh
 
-ssh -i ${private_key} ${instance_hostname} "${@}"
+# keep alive 60sec and timeout after 30 tries
+ssh -o ServerAliveInterval=60 -o ServerAliveCountMax=30 -i ${private_key} ${instance_hostname} "${@}"
```
cdesiniotis, line 8: Question -- why is this required?

tests/scripts/verify-operator.sh
cdesiniotis: Why do we need a while loop and all of this extra logic? Running `kubectl --wait` for both the driver and validator pod should be all we need, and we can exit when a timeout is reached.

Can we simplify this entire file to just:
```
kubectl --wait nvidia-driver-daemonset ...
kubectl --wait nvidia-operator-validator ...

if rc != 0 {
    collect_logs
}
```

And we can extract the `collect_logs` function to its own script.

tests/scripts/verify-operator.sh
```diff
@@ -11,9 +11,42 @@ source ${SCRIPT_DIR}/.definitions.sh
...
+    fi
+
+    check_pod_ready "nvidia-driver-daemonset" ${DAEMON_POD_STATUS_TIME_OUT} && \
+        check_pod_ready "nvidia-operator-validator" ${POD_STATUS_TIME_OUT}; status=$?
+    
+    if [ $status -ne 0 ]; then
+        not_ready_pod_status=$(kubectl get pods -n $TEST_NAMESPACE -o jsonpath="{.items[?(@.status.phase != 'Running')].metadata.name}")
+        if [ -n "$not_ready_pod_status" ]; then
+            for pod in $not_ready_pod_status; do
+                echo "Collecting logs for pod: $pod in dir ${log_dir}"
```
cdesiniotis, line 38: As mentioned in a previous comment, lets extract this into a separate function, `collect_logs`, which we can define in a separate file.

Note, we have a script named `must-gather.sh` which we ask users to run when they are hitting operator issues. It collects a comprehensive set of logs to help us debug. Feel free to reuse some of the logic from there, specifically the logic to collect pod logs: https://github.com/NVIDIA/gpu-operator/blob/main/hack/must-gather.sh#L169-L196. If it makes sense, we can consider just running the `must-gather.sh` script directly from within our tests to collect logs.

cdesiniotis: Also, I downloaded the logs generated from this failing pipeline, and all of the files are empty: https://github.com/shivakunv/gpu-driver-container/actions/runs/10425570121
Can you check this?

tests/scripts/checks.sh
shivakunv: uninstall-operator.sh  created . 
helm uninstall $release_name --namespace $namespace was already there at the end of the function . 
_Uninstalling Helm itself will not remove DaemonSets or Secrets_

tests/scripts/remote.sh
shivakunv: default ssh server timeout 60sec
_kubectl wait or sleep  :
 ssh might drop the connection if there's no activity between the client and server ( default timeout 60sec)_

tests/scripts/verify-operator.sh
shivakunv: I removed the while loop and added more delay since the gpu-operator sometimes takes longer to deploy. 
increased the daemon wait time to 15 minutes.

I kept the must-gather.sh script as it is, as it only requires the artifacts directory path for log collection. 
I have integrated it with the current code

shivakunv:
@cdesiniotis PTAL

.github/workflows/ci.yaml
```diff
@@ -21,15 +21,10 @@ on:
       - completed
     branches:
       - main
-
+
```
cdesiniotis, line 24: nit: remove this unneeded change.

tests/scripts/must-gather.sh
```diff
@@ -0,0 +1,256 @@
+#!/usr/bin/env bash
```
cdesiniotis, line 1: Instead of maintaining a copy of this script here, let's download this script on the remote machine at runtime.

tests/scripts/uninstall-operator.sh
```diff
@@ -0,0 +1,18 @@
...
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source ${SCRIPT_DIR}/.definitions.sh
+
+namespace=$1
+release_name=$2
+kubectl delete daemonsets --all -n $namespace || true
+kubectl delete deployments --all -n $namespace || true
+kubectl delete services --all -n $namespace || true
+kubectl delete pods --all -n $namespace || true
+kubectl delete namespace $namespace || true
```
cdesiniotis, line 17: These `kubectl delete` commands are not needed, as the helm uninstall should delete all resources in the namespace. If desired, you can delete the namespace after uninstalling the helm chart.

tests/scripts/must-gather.sh
shivakunv: added

cdesiniotis (approved):

PR #93, opened 2024-08-22
end-to-end  log_dir fix

shivakunv:
@cdesiniotis immediate fix  for main branch e2e failure

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,7 +72,7 @@ jobs:
           fi
         done
         source ./tests/scripts/.definitions.sh
-        ./tests/scripts/pull.sh ${LOG_DIR} logs
+        ./tests/scripts/pull.sh /tmp/logs logs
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
           fi
         done
         source ./tests/scripts/.definitions.sh
-        ./tests/scripts/pull.sh ${LOG_DIR} logs
+        ./tests/scripts/pull.sh /tmp/logs logs
```
cdesiniotis, line 75: Question -- if we reach this point, will there be any logs at `/tmp/logs`? My understanding is we only collect logs, by running `must-gather.sh`, only when a test case fails.

shivakunv: Just to satisfy the artifacts step, if I put it inside an if condition, the artifacts step will fail with no path : ./logs

PR #126, opened 2024-09-20
Update driver-container for SUSE SLE

e4t:
This set of patches fixes the container for SUSE SLE and incorporates some of the updates to the containers for other distros.


rahulait:
Given that this PR is almost 2 years old now and we don't maintain sle15 related code, I am going to close this PR. If required in future, we can re-open the PR

PR #141, opened 2024-10-15
pass golang version as docker build arg and bump version to 1.22.8

tariq1890:


cdesiniotis:
We should define GOLANG_VERSION in one place like we do in other projects.

tariq1890:
@cdesiniotis Done, please check now

cdesiniotis (approved):

PR #144, opened 2024-10-22
bump R550 and R535 to the latest TRD releases

tariq1890:
[550.127.05](https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-550-127-05/index.html) and [535.216.01](https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-535-216-01/index.html) releases are now live.


cdesiniotis (approved):
Should we remove R560 as I don't believe it will receive any more updates?

tariq1890:
For sure, I can do that in a follow up PR. I also need to clean up the CI pipeline yaml file to remove some unsupported distros

tariq1890:
> Should we remove R560 as I don't believe it will receive any more updates?

@cdesiniotis done

PR #151, opened 2024-10-29
add gpu driver version 565.57.01

tariq1890:


.github/workflows/image.yaml
```diff
@@ -49,6 +50,9 @@ jobs:
           - ispr: true
             dist: ubuntu20.04
             driver: 550.127.05
+          - ispr: true
+            dist: ubuntu20.04
+            driver: 565.57.01
```
cdesiniotis, line 55: Question -- are we able to replace the three exclude bullets with just
```
- ispr: true
  dist: ubuntu20.04
```
since we are effectively skipping ubuntu20.04 on PRs?

tariq1890: tbh, I just added it for the sake of adding an excludes directive. I don't think we need to exclude add ubuntu20.04 dists

cdesiniotis (approved):

PR #172, opened 2024-12-11
[ubuntu24.04] grant _apt user root privileges

tariq1890:
This PR fixes the following warning observed when running the ubuntu24.04 gpu driver container

`W: Download is performed unsandboxed as root as file '/tmp/tmp.t07XczEtx1/linux-image-6.8.0-47-generic_6.8.0-47.47_amd64.deb' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)`

This is consistent with the driver containers for the other OS versions


cdesiniotis:
Can you point me to where we do this in other distros?

tariq1890:
[Here](https://github.com/NVIDIA/gpu-driver-container/blob/main/ubuntu22.04/install.sh#L43) and [here](https://github.com/NVIDIA/gpu-driver-container/blob/main/ubuntu20.04/install.sh#L43).

cdesiniotis (approved):

PR #175, opened 2024-12-12
fix precompiled driver container failures when enabling OpenRM

tariq1890:
We have found that the OpenRM modules fail to load successfully as it's unable to find the firmware files in the `/lib/firmware` directory. This was happening due to two reasons:

1. The `/lib/firmware` files were being installed during the container image build-time. See [here](https://github.com/NVIDIA/gpu-driver-container/blob/aca3aca650f3a6c3d465d7bb6606b2c074abde76/ubuntu22.04/precompiled/Dockerfile#L52)
2. The `/lib/firmware` files were wiped out as we mount the `/run/nvidia/driver/lib/firmware` directory . See [here](https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-driver/0500_daemonset.yaml#L133-L134)

To fix this issue, we make the following changes

1. We only download (and not install) the `nvidia-driver-${DRIVER_BRANCH}-server` package, so that the actual takes place during container runtime and the `lib/firmware` files are installed AFTER the mount
2. Add logic to point the override firmware file search path in `/sys/module/firmware_class/parameters/path`. This is performed in non precompiled containers and it is needed for the OpenRM Kernel module to load successfully

### DIFF
--- a/ubuntu22.04/precompiled/nvidia-driver
+++ b/ubuntu22.04/precompiled/nvidia-driver
@@ -101,6 +101,25 @@ _load_driver() {
     echo "Parsing kernel module parameters..."
     _get_module_params
 
+    local nv_fw_search_path="$RUN_DIR/driver/lib/firmware"
+    local set_fw_path="true"
+    local fw_path_config_file="/sys/module/firmware_class/parameters/path"
+    for param in "${NVIDIA_MODULE_PARAMS[@]}"; do
+        if [[ "$param" == "NVreg_EnableGpuFirmware=0" ]]; then
+          set_fw_path="false"
+        fi
+    done
+
+    if [[ "$set_fw_path" == "true" ]]; then
+        echo "Configuring the following firmware search path in '$fw_path_config_file': $nv_fw_search_path"
+        if [[ ! -z $(grep '[^[:space:]]' $fw_path_config_file) ]]; then
+            echo "WARNING: A search path is already configured in $fw_path_config_file"
+            echo "         Retaining the current configuration"
+        else
+            echo -n "$nv_fw_search_path" > $fw_path_config_file || echo "WARNING: Failed to configure firmware search path"
+        fi
+    fi
+
     echo "Loading ipmi and i2c_core kernel modules..."
     modprobe -a i2c_core ipmi_msghandler ipmi_devintf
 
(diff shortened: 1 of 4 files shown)


ubuntu22.04/precompiled/nvidia-driver
```diff
@@ -124,6 +143,10 @@ _load_driver() {
 
     DRIVER_VERSION=$(nvidia-smi -q | grep "Driver Version" | awk -F: '{print $2}' | xargs)
     if _assert_nvswitch_system; then
+        echo "Installing NVIDIA fabric manager and libnvidia NSCQ packages..."
+        apt-get install -y --no-install-recommends \
+            nvidia-fabricmanager-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 \
+            libnvidia-nscq-${DRIVER_BRANCH}=${DRIVER_VERSION}-1
```
cdesiniotis, line 149: Question -- should this be in `_install_driver` instead?

tariq1890: I thought it's better to install it here since we check for n switch devices and we also grab the DRIVER_VERSION from nvidia-smi in this method

tariq1890: > I understand moving the installation of fabricmanager / nscq to _install_driver() would result us in calling _assert_nvswitch_system() twice

We are actually only calling it once. Right now, the _assert_nvswitch_system() was only used to check whether to start nv-fabricmanager or not, now we are also installing the downloaded fabricmanager and nscq packages

> Isn't DRIVER_VERSION already set as an ENV in the Dockerfile? So why do we even need to resolve it on line 144? 

Correct,  I just thought the `nvidia-smi` output is more reliable as it's fetching the actual version of the driver installed via the precompiled package. Moreover, we already do this in the ubuntu20.04 implementation. See [here](https://github.com/NVIDIA/gpu-driver-container/blob/84bff94ceb321bb9e75ec49bd95f309ece1caf74/ubuntu20.04/precompiled/nvidia-driver#L126)

cdesiniotis: Got it. I am fine with this implementation then.

cdesiniotis (approved):

ubuntu22.04/precompiled/nvidia-driver
tariq1890: Thanks!

PR #176, opened 2024-12-12
[ubuntu24.04] remove usage of deprecated 'apt-key add'

cdesiniotis:


ubuntu24.04/Dockerfile
```diff
@@ -64,8 +64,11 @@ RUN echo "TARGETARCH=$TARGETARCH"
 ADD install.sh /tmp
 
 # Fetch GPG keys for CUDA repo
-RUN apt-key del 7fa2af80 && OS_ARCH=${TARGETARCH/amd64/x86_64} && OS_ARCH=${OS_ARCH/arm64/sbsa} && \
-    apt-key adv --fetch-keys "https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/${OS_ARCH}/3bf863cc.pub"
+RUN apt-key del 7fa2af80 3bf863cc && \
```
tariq1890, line 67: Do we know if these GPG keys are already added to the container image?

If not, we could just remove it and add them when it becomes necessary

ubuntu24.04/Dockerfile
```diff
@@ -64,8 +64,11 @@ RUN echo "TARGETARCH=$TARGETARCH"
...
+    apt-get update && apt-get install -y --no-install-recommends curl gpg && \
+    OS_ARCH=${TARGETARCH/amd64/x86_64} && OS_ARCH=${OS_ARCH/arm64/sbsa} && \
+    curl -fsSL https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/${OS_ARCH}/3bf863cc.pub | gpg --dearmor -o /etc/apt/keyrings/cuda-3bf863cc.pub && \
+    echo "deb [signed-by=/etc/apt/keyrings/cuda-3bf863cc.pub] https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/${OS_ARCH} /" > /etc/apt/sources.list.d/cuda.list
```
tariq1890, line 71: Just a nit, but why not this? Just looks a tad better IMO
```suggestion
    echo "deb [signed-by=/etc/apt/keyrings/cuda.pub] https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/${OS_ARCH} /" > /etc/apt/sources.list.d/cuda.list
```

ubuntu24.04/Dockerfile
cdesiniotis: 3bf863cc is included. 7fa2af80 is the old one and it is not in the ubuntu 24.04 image. I can omit 7fa2af80 from this command.

ubuntu24.04/Dockerfile
cdesiniotis: Updated to just `cuda.pub`

ubuntu24.04/Dockerfile
```diff
@@ -64,8 +64,12 @@ RUN echo "TARGETARCH=$TARGETARCH"
 ADD install.sh /tmp
 
 # Fetch GPG keys for CUDA repo
-RUN apt-key del 7fa2af80 && OS_ARCH=${TARGETARCH/amd64/x86_64} && OS_ARCH=${OS_ARCH/arm64/sbsa} && \
-    apt-key adv --fetch-keys "https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/${OS_ARCH}/3bf863cc.pub"
+RUN apt-key del 3bf863cc && \
```
tariq1890, line 67: Looks like this statement is running in the final layer, can we `apt purge` the curl package after performing this action?

tariq1890: Let's also add this line as it helps shrink the image size further

```
rm -rf /var/lib/apt/lists/*; \
```

tariq1890 (approved):

tariq1890 (approved):

PR #179, opened 2025-01-03
add go-based script to auto-select kernel module type

tariq1890:


driver-assistant/src/main.go
```diff
@@ -0,0 +1,236 @@
...
+	// DefaultSupportedGpusJsonPath represents the default install location of the supported-gpus.json file
+	DefaultSupportedGpusJsonPath = "/usr/share/nvidia-driver-assistant/supported-gpus/supported-gpus.json"
+
+	// DriverHintUnknown is used when the gpu device is not found in supported-gpus.json
+	DriverHintUnknown = "unknown"
+	// DriverHintOpenRequired is used when the gpu device compulsorily needs the OpenRM kernel modules
+	DriverHintOpenRequired = "open-required"
+	// DriverHintProprietaryRequired is used when the gpu device compulsorily needs the proprietary kernel modules
+	DriverHintProprietaryRequired = "proprietary-required"
+	//DriverHintAny is used when the gpu device can support either kernel module types
```
cdesiniotis, line 35: nit
```suggestion
	// DriverHintAny is used when the gpu device can support either kernel module types
```

driver-assistant/src/main.go
```diff
@@ -0,0 +1,236 @@
...
+	supportedGpusJsonPath string
+	driverBranch          int
+)
+
+type GPUDevice struct {
+	ID           string   `json:"devid"`
+	Name         string   `json:"name"`
+	LegacyBranch string   `json:"legacybranch"`
+	Features     []string `json:"features"`
+}
```
cdesiniotis, line 54: nit: newline
```suggestion
}

```

driver-assistant/src/main.go
```diff
@@ -0,0 +1,236 @@
...
+	selectKernelModule := cli.Command{}
+	selectKernelModule.Name = "select-kernel-module"
+	selectKernelModule.Usage = "Auto-select the kernel module based on the detected GPUs"
+	selectKernelModule.Action = func(c *cli.Context) error {
+		return SelectKernelModule(c)
+	}
+
+	selectKernelModuleFlags := []cli.Flag{
+		&cli.StringFlag{
+			Name:        "supported-gpus",
```
cdesiniotis, line 81: nit: let's rename this flag to indicate it is a file path?
```suggestion
			Name:        "supported-gpus-file",
```

driver-assistant/src/main.go
```diff
@@ -0,0 +1,236 @@
...
+
+func buildGPUSearchMap(data GPUData) map[string]GPUDevice {
+	var gpuMap = make(map[string]GPUDevice)
+	for _, gpuDevice := range data.Chips {
+		gpuMap[gpuDevice.ID] = gpuDevice
+	}
+	return gpuMap
+}
+
+func getNvidiaGPUs() ([]string, error) {
```
cdesiniotis, line 205: Rather than reimplement this, let's use the APIs from go-nvlib: https://github.com/NVIDIA/go-nvlib/blob/main/pkg/nvpci/nvpci.go

driver-assistant/src/main.go
```diff
@@ -0,0 +1,236 @@
...
+	driverHints := getDriverHints(gpuDevices, searchMap)
+	if slices.Contains(driverHints, DriverHintOpenRequired) &&
+		slices.Contains(driverHints, DriverHintProprietaryRequired) {
+		return "", fmt.Errorf("unsupported GPU topology")
+	} else if slices.Contains(driverHints, DriverHintOpenRequired) {
+		return "kernel-open", nil
+	} else if slices.Contains(gpuDevices, DriverHintProprietaryRequired) {
+		return "kernel", nil
+	} else {
+		return getDriverBranchDefault(driverBranch), nil
```
cdesiniotis, line 165: Question: If we reach this point, I assume the `driverHints` slice only contains `DriverHintUnknown` meaning that the GPU device ID was not found in the supported-gpus.json file. I am guessing this would probably only occur on a system with newer GPUs where the json file is stale. Should we depend on `getDriverBranchDefault(driverBranch)` in this case or always return the open modules -- since we know that newer GPUs will always support the open modules? 

My question may not be super relevant, since I would expect that newer GPUs would require installing a newer driver (>=560) anyways...

driver-assistant/src/main.go
tariq1890: thanks

driver-assistant/src/main.go
tariq1890: agreed

driver-assistant/src/main.go
tariq1890: > Question: If we reach this point, I assume the driverHints slice only contains DriverHintUnknown meaning that the GPU device ID was not found in the supported-gpus.json file.

`DriverHintAny` as well

It's a fair question. The `nvidia-driver-assistant` follows that logic, I guess if we are certain all GPUs starting from Blackwell are not going to support proprietary modules, we could always default to OpenRM regardless of the driver branch.  I am ambivalent to this, I don't think the current behaviour isn't too bad either.

driver-assistant/src/main.go
tariq1890: Normally I'd agree, but my preference here is to keep the dep tree as minimal as possible. Importing `go-nvlib` will add more transitive deps

driver-assistant/src/main.go
cdesiniotis: For `DriverHintAny`, we should depend on the default for the driver branch. I guess I am questioning what the behavior should be for `DriverHintUnknown`. I think the current behavior -- depending on the default for the driver branch -- is okay.

driver-assistant/src/main.go
cdesiniotis: That is fair.

driver-assistant/src/main.go
tariq1890: I guess I am not too opposed to returning `kernel-open` for unknown gpu devices

cdesiniotis: Yeah, the logic being that all new SKUs *should* support the open kernel modules. So it should be safe for this to be the default.

tariq1890: Updated the PR. Please check now

cdesiniotis: LGTM

cdesiniotis (approved):

PR #181, opened 2025-01-08
check-in and embed latest supported-gpus.json file

tariq1890:
Since we don't an authoritative source to retrieve the latest `supported-gpus.json`, we work around this by embedding the supported-gpus.json file into the `gpu-driver-util` application.

Once a publicly hosted link of supported-gpus.json, we can remove this dependency.

NOTE: The supported-gpus.json checked into this branch was retrieved from the `nvidia-driver-assistant-0.9.57.01-1` deb package


gpu-driver-util/main.go
```diff
@@ -177,12 +181,14 @@ func resolveKernelModuleType(gpuDevices []string, searchMap map[string]GPUDevice
 	if requiresOpenRM && requiresProprietary {
 		return "", fmt.Errorf("unsupported GPU topology")
 	} else if requiresOpenRM {
-		return KernelModuleTypeOpen, nil
+		kernelModuleType = KernelModuleTypeOpen
 	} else if requiresProprietary {
+		kernelModuleType = KernelModuleTypeProprietary
```
cdesiniotis, line 186: Did you mean to remove the `return` statement after this line?

tariq1890: ah yes, good catch. Will remove it

cdesiniotis (approved):

gpu-driver-util/main.go
tariq1890: Please check now

PR #188, opened 2025-01-14
Make CUDA version configurable for RHEL9, default to 12.6.2

empovit:


tariq1890:
Hi @empovit , thanks for the PR. We previously had variable substitution for the  cuda base image version before, but we changed that as we needed to use @dependabot to auto-bump the container image versions

empovit:
Thanks @tariq1890 for the explanation. It makes sense, but it also means that https://github.com/NVIDIA/gpu-driver-container/blob/a5483c34513d750ceefda4f9cc1a015ea9f288fb/rhel9/precompiled/Makefile#L57 doesn't work. Am I mistaken?

empovit:
BTW, how does it work with the `CUDA_DIST` variable in `FROM nvcr.io/nvidia/cuda:12.6.2-base-${CUDA_DIST}`? Is dependabot smart enough to understand just the first part?

tariq1890:
> doesn't work. Am I mistaken?

You're right that it doesn't work. We should probably remove that since it's misleading

> BTW, how does it work with the CUDA_DIST variable in FROM nvcr.io/nvidia/cuda:12.6.2-base-${CUDA_DIST}? Is dependabot smart enough to understand and bump just the first part?

From what I know about dependabot, I don't think that will work. I know this is something that the renovate bot can do, but I am not sure if depedabot has been improved to have feature parity with the renovate bot

empovit:
Sorry I'm confused. You're saying that my fix isn't good because the existing behavior is intentional - so that dependabot could auto-bump the container image versions. Yet at the same time you're saying that dependabot can't currently do it because the `FROM` line already has variable substitution, namely `${CUDA_DIST}`.

tariq1890:
Note that the `rhel9` folder isn't one we actively maintain yet. We plan on doing that soon however.

To get dependabot to work we need to remove the ${CUDA_DIST} variable substitution as well. I hope that clarifies

empovit:
@tariq1890 tale a look at https://docs.renovatebot.com/. This bot is very flexible in updating image tags, and makes it possible to update `ARG CUDA_VERSION='12.6.2'`, e.g. whenever the CUDA  image gets a new tag, while keeping the variable substitution in `FROM nvcr.io/nvidia/cuda:${CUDA_VERSION}-base-${CUDA_DIST}`.

PR #192, opened 2025-01-16
Reintroduce OPEN_KERNEL_MODULES_ENABLED envvar for backwards compatib…

cdesiniotis:
…ility

### DIFF
--- a/ubuntu22.04/nvidia-driver
+++ b/ubuntu22.04/nvidia-driver
@@ -15,7 +15,9 @@ NVIDIA_UVM_MODULE_PARAMS=()
 NVIDIA_MODESET_MODULE_PARAMS=()
 NVIDIA_PEERMEM_MODULE_PARAMS=()
 TARGETARCH=${TARGETARCH:?"Missing TARGETARCH env"}
-KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE:-auto}
+
+OPEN_KERNEL_MODULES_ENABLED=${OPEN_KERNEL_MODULES_ENABLED:-}
+KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE:-}
 
 export DEBIAN_FRONTEND=noninteractive
 
@@ -480,6 +482,13 @@ _shutdown() {
 }
 
 _resolve_kernel_type() {
+  # For backwards compatibility with older GPU Operator versions where KERNEL_MODULE_TYPE is not set,
+  # honor the deprecated OPEN_KERNEL_MODULES_ENABLED field
+  if [ -z "${KERNEL_MODULE_TYPE}" ]; then
+    [[ "${OPEN_KERNEL_MODULES_ENABLED}" == "true" ]] && KERNEL_TYPE=kernel-open || KERNEL_TYPE=kernel
+    return 0
+  fi
+
   if [ "${KERNEL_MODULE_TYPE}" == "proprietary" ]; then
     KERNEL_TYPE=kernel
   elif [ "${KERNEL_MODULE_TYPE}" == "open" ]; then
--- a/ubuntu22.04/nvidia-driver
+++ b/ubuntu22.04/nvidia-driver
@@ -15,9 +15,7 @@ NVIDIA_UVM_MODULE_PARAMS=()
 NVIDIA_MODESET_MODULE_PARAMS=()
 NVIDIA_PEERMEM_MODULE_PARAMS=()
 TARGETARCH=${TARGETARCH:?"Missing TARGETARCH env"}
-
-OPEN_KERNEL_MODULES_ENABLED=${OPEN_KERNEL_MODULES_ENABLED:-false}
-[[ "${OPEN_KERNEL_MODULES_ENABLED}" == "true" ]] && KERNEL_TYPE=kernel-open || KERNEL_TYPE=kernel
+KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE:-auto}
 
 export DEBIAN_FRONTEND=noninteractive
 
@@ -481,6 +479,24 @@ _shutdown() {
     return 1
 }
 
+_resolve_kernel_type() {
+  if [ "${KERNEL_MODULE_TYPE}" == "proprietary" ]; then
+    KERNEL_TYPE=kernel
+  elif [ "${KERNEL_MODULE_TYPE}" == "open" ]; then
+    KERNEL_TYPE=kernel-open
+  elif [ "${KERNEL_MODULE_TYPE}" == "auto" ]; then
+      kernel_module_type=$(/drivers/NVIDIA-Linux-$DRIVER_ARCH-$DRIVER_VERSION/nvidia-installer --print-recommended-kernel-module-type)
+      if [ $? -ne 0 ]; then
+          [[ "${DRIVER_BRANCH}" -lt 560 ]] && kernel_module_type=proprietary || kernel_module_type=open
+          echo "failed to retrieve the recommended kernel module type from nvidia-installer, falling back to the '$kernel_module_type' driver"
+      fi
+      [[ "${kernel_module_type}" == "open" ]] && KERNEL_TYPE=kernel-open || KERNEL_TYPE=kernel
+  else
+    echo "invalid value for the KERNEL_MODULE_TYPE variable: ${KERNEL_MODULE_TYPE}"
+    return 1
+  fi
+}
+
 _find_vgpu_driver_version() {
     local count=""
     local version=""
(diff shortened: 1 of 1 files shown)


ubuntu22.04/nvidia-driver
```diff
@@ -15,6 +15,8 @@ NVIDIA_UVM_MODULE_PARAMS=()
 NVIDIA_MODESET_MODULE_PARAMS=()
 NVIDIA_PEERMEM_MODULE_PARAMS=()
 TARGETARCH=${TARGETARCH:?"Missing TARGETARCH env"}
+
+OPEN_KERNEL_MODULES_ENABLED=${OPEN_KERNEL_MODULES_ENABLED:-""}
```
tariq1890, line 19: Does this statement lead to this variable being set at all times?

cdesiniotis: It will be set to the empty string by default. I can update this if required. I have confirmed though that `[ -n ${OPEN_KERNEL_MODULES_ENABLED} ]` return true if the variable is unset or is the empty string.

tariq1890: Thanks

ubuntu22.04/nvidia-driver
```diff
@@ -480,6 +482,14 @@ _shutdown() {
 }
 
 _resolve_kernel_type() {
+
+  # For backwards compatibility with older GPU Operator versions, honor the OPEN_KERNEL_MODULES_ENABLED
+  # variable when KERNEL_MODULE_TYPE is not set.
+  if [ -n ${OPEN_KERNEL_MODULES_ENABLED} ] && [ -z "${KERNEL_MODULE_TYPE}" ]; then
```
tariq1890, line 488: In L20 we default the KERNEL_MODULE_TYPE to auto. Are we sure that this conditional will evaluate to true ?

cdesiniotis: It won't, you are right. What about the following --  
1) do not set a default value for KERNEL_MODULE_TYPE in the driver container
2) GPU Operator 24.12.0+ should will always set the KERNEL_MODULE_TYPE envvar. Even if driver.kernelModuleType is not set in ClusterPolicy, the operator will set the envvar to `auto`.

tariq1890: Yeah that sounds good to me

ubuntu22.04/nvidia-driver
```diff
@@ -480,6 +482,22 @@ _shutdown() {
...
+  if [ -z "${KERNEL_MODULE_TYPE}" ]; then
+   if [ ${OPEN_KERNEL_MODULES_ENABLED} == "true" ]; then
+     echo "Honoring deprecated field OPEN_KERNEL_MODULES_ENABLED=${OPEN_KERNEL_MODULES_ENABLED}"
+     KERNEL_TYPE=kernel-open
+     return 0
+   fi
+
+   # If both KERNEL_MODULE_TYPE and OPEN_KERNEL_MODLES_ENABLED are unset, default
+   # to the open kernel modules on newer driver branches
+   [[ "${DRIVER_BRANCH}" -gt "560" ]] && KERNEL_TYPE=kernel-open || KERNEL_TYPE=kernel
```
tariq1890, line 497: What about the edge cases ? Drivers that only support one type of kernel modules?

ubuntu22.04/nvidia-driver
```diff
@@ -480,12 +482,32 @@ _shutdown() {
...
+  if [ -z "${KERNEL_MODULE_TYPE}" ]; then
+   if [ ${OPEN_KERNEL_MODULES_ENABLED} == "true" ]; then
+     echo "Honoring deprecated field OPEN_KERNEL_MODULES_ENABLED=${OPEN_KERNEL_MODULES_ENABLED}"
+     KERNEL_TYPE=kernel-open
+     return 0
+   fi
+
+   # If neither KERNEL_MODULE_TYPE or OPEN_KERNEL_MODULES_ENABLED are set, get the default
+   # kernel module type from the nvidia-installer.
+   KERNEL_TYPE=$(nvidia-installer --print-recommended-kernel-module-type)
```
tariq1890, line 497: note that this command `nvidia-installer --print-recommended-kernel-module-type` returns either `open` or `proprietary`. 

i) If it's `open`, we set `KERNEL_TYPE=kernel-open`
ii) If it's `proprietary`, we set `KERNEL_TYPE=kernel`

tariq1890:
Thanks for creating this @cdesiniotis . Please rebase this on top of the latest main

ubuntu22.04/nvidia-driver
cdesiniotis: Thanks for pointing this out. Will update this in my next iteration.

cdesiniotis: Actually, I will remove this call altogether as we agreed to maintain the old behavior, we would default to proprietary in this case.

ubuntu22.04/nvidia-driver
```diff
@@ -481,6 +481,32 @@ _shutdown() {
...
+
+  if [ "${KERNEL_MODULE_TYPE}" == "proprietary" ]; then
+    KERNEL_TYPE=kernel
+  elif [ "${KERNEL_MODULE_TYPE}" == "open" ]; then
+    KERNEL_TYPE=kernel-open
+  elif [ "${KERNEL_MODULE_TYPE}" == "auto" ]; then
+      kernel_module_type=$(nvidia-installer --print-recommended-kernel-module-type)
+      if [ $? -ne 0 ]; then
+          echo "cannot autodetect the kernel module type, printing error logs from /var/log/gpu-driver-util.log..."
+          tail -n 3 /var/log/gpu-driver-util.log
```
tariq1890, line 500:
```suggestion
          echo "failed to retrieve the recommended kernel module type from nvidia-installer"
```

cdesiniotis: Thanks. Fixed.

ubuntu22.04/nvidia-driver
```diff
@@ -481,6 +481,31 @@ _shutdown() {
...
+    [[ "${OPEN_KERNEL_MODULES_ENABLED}" == "true" ]] && KERNEL_TYPE=kernel-open || KERNEL_TYPE=kernel
+    return 0
+  fi
+
+  if [ "${KERNEL_MODULE_TYPE}" == "proprietary" ]; then
+    KERNEL_TYPE=kernel
+  elif [ "${KERNEL_MODULE_TYPE}" == "open" ]; then
+    KERNEL_TYPE=kernel-open
+  elif [ "${KERNEL_MODULE_TYPE}" == "auto" ]; then
+      kernel_module_type=$(nvidia-installer --print-recommended-kernel-module-type)
```
tariq1890, line 497: Shall we move the `nvidia-installer` to the `PATH` during build-time?

ubuntu22.04/nvidia-driver
```diff
@@ -481,6 +481,31 @@ _shutdown() {
...
+      if [ $? -ne 0 ]; then
+          echo "failed to retrieve the recommended kernel module type from nvidia-installer"
```
tariq1890, line 499: thinking out loud here. We will have essentially added a point of failure here. 

What are your thoughts on a lazy driver-branch-based logic as a fallback? If driver branch >= 560, we return `open`, otherwise we return `proprietary`

ubuntu22.04/nvidia-driver
cdesiniotis: Possibly. That would require us to extract the runfile during build time, which we aren't doing currently.

ubuntu22.04/nvidia-driver
cdesiniotis: I think I am generally okay with it. It is an extra safeguard in case the nvidia-installer CLI has bug.

ubuntu22.04/nvidia-driver
tariq1890: Fair enough; we probably shouldn't extract it during build-time, but we will still need to move the `nvidia-installer` to the PATH

cdesiniotis: Yep. Extracting at build time will increase our image size significantly. We will need to move it to the path or `cd` into the local directory before invoking `_resolve_kernel_type` (like we do elsewhere).

tariq1890: or cd into `NVIDIA-Linux-$DRIVER_ARCH-$DRIVER_VERSION`

cdesiniotis: That assumes the runfile has been extracted beforehand though.

tariq1890: > Yep. Extracting at build time will increase our image size significantly. We will need to move it to the path or cd into the local directory before invoking _resolve_kernel_type (like we do elsewhere).

Ah, just seeing this comment now. Sounds good

ubuntu22.04/nvidia-driver
cdesiniotis: Updated as you have suggested.

ubuntu22.04/nvidia-driver
cdesiniotis: I pushed an update, which uses the absolute path to the `nvidia-installer`. It assumes the runfile has been extracted prior to the call to `_resolve_kernel_type`.

cdesiniotis:
Closing as this was implemented in https://github.com/NVIDIA/gpu-driver-container/pull/199

PR #193, opened 2025-01-17
Revert gpu driver util commits

tariq1890:


tariq1890:
> We are reverting all of these commits because when we re-introduce the auto-detection, we will use the nvidia-installer rather than our own gpu-driver-util, right?

Correct, that is the plan

PR #222, opened 2025-02-11
[ubuntu-precompiled] enable offline installation of driver packages

tariq1890:
This PR introduces a completely new method of installing the nvidia precompiled driver packages. 

## Motivation 
We recently discovered that our offline installs weren't exactly "offline". During the driver container run-time, the driver container was still downloading packages externally. The root cause of this was the multiple `apt-get update` calls, which would erase the previously downloaded packages from the `apt-get install -y --no-install-recommends --download-only <packages>` command.

## Summary of changes:
i) Download the packages and its dependencies using the following commands
```
apt-get download $PACKAGE
apt-get download $(apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances $PACKAGE | grep "^\w" | sort -u)
```
ii) Move to a permanent location and gzip the apt-get downloaded packages.
iii) Create a local apt package source pointing to the new directory with the downloaded packages

NOTE: In this PR, I also move away from installing the giant metapackage `nvidia-drivers-${DRIVER_VERSION}-server` and purging unneeded packages thereafter. Instead, we just install the packages that we actually need instead of worrying about any bloat

### DIFF
--- a/ubuntu22.04/precompiled/Dockerfile
+++ b/ubuntu22.04/precompiled/Dockerfile
@@ -48,27 +48,20 @@ RUN if [ -n "${CVE_UPDATES}" ]; then \
         rm -rf /var/lib/apt/lists/*; \
     fi
 
-# update pkg cache and install pkgs for userspace driver libs
-RUN apt-get update && apt-get install -y --download-only --no-install-recommends nvidia-driver-${DRIVER_BRANCH}-server \
-    nvidia-fabricmanager-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 \
-    libnvidia-nscq-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 && \
-    rm -rf /var/lib/apt/lists/*;
-
 RUN if [ "$DRIVER_BRANCH" -ge "550" ]; then \
     apt-get update && \
     apt-get install -y --no-install-recommends nvidia-imex-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 && \
     rm -rf /var/lib/apt/lists/*; fi
 
-# update pkg cache and download pkgs for driver module installation during runtime.
-# this is done to avoid shipping .ko files.
-# avoid cleaning the cache after this to retain these packages during runtime.
-RUN apt-get update && apt-get install --download-only --no-install-recommends -y linux-objects-nvidia-${DRIVER_BRANCH}-server-${KERNEL_VERSION} \
-    linux-signatures-nvidia-${KERNEL_VERSION} \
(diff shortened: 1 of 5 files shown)


ubuntu22.04/precompiled/Dockerfile
```diff
@@ -48,28 +48,21 @@ RUN if [ -n "${CVE_UPDATES}" ]; then \
         rm -rf /var/lib/apt/lists/*; \
     fi
 
-# update pkg cache and install pkgs for userspace driver libs
-RUN apt-get update && apt-get install -y --download-only --no-install-recommends nvidia-driver-${DRIVER_BRANCH}-server \
```
shivamerla, line 52: Where else was `apt-get remote or apt-get clean` run after this step which was causing the packages to be removed? `apt-get update` itself doesn't remove downloaded packages.

ubuntu22.04/precompiled/nvidia-driver
```diff
@@ -234,15 +234,11 @@ _unload_driver() {
...
-    apt-get install -y --no-install-recommends nvidia-driver-${DRIVER_BRANCH}-server
-
-    # Uninstall unnecessary packages installed as a part of the nvidia-driver-${DRIVER_BRANCH}-server package
-    apt-get purge -y \
-        libnvidia-egl-wayland1 \
-        nvidia-dkms-${DRIVER_BRANCH}-server \
-        nvidia-kernel-source-${DRIVER_BRANCH}-server \
-        xserver-xorg-video-nvidia-${DRIVER_BRANCH}-server
+    # Install necessary driver userspace packages
+    apt-get install -y --no-install-recommends \
```
shivamerla, line 238: These will just install utils like `nvidia-smi, nvidia-persistenced, nvidia-mps-control` etc, but not other required libs like encoder/decoder/gl/cuda libs etc. There are ton of other user space packages that we need. where are those installed as dependencies?

ubuntu22.04/precompiled/Dockerfile
tariq1890: So when I downloaded packages using `apt-get install --download-only` and then ran an update `apt-get update`. I noticed that the downloaded packages in `/var/cache/apt/archives` were wiped out

ubuntu22.04/precompiled/nvidia-driver
tariq1890: If the `encoder/decoder/gl/cuda libs` are required, I can install them. I wasn't aware that they were needed. 

I validated the current container with mps and tests like gpu-burn. They worked fine

tariq1890: I've added the other packages. Please check again

cdesiniotis (approved):

PR #228, opened 2025-02-13
[no-relnote][ci-image-build] pin tonistiigi/binfmt to v6.2.0

shivakunv:


elezar:
@shivakunv the issue now seems to be that we're missing the correct IMEX packages for 550 rhel8. Could it be that we shouldn't be trying to install this package there? 

```
8.161 Error: Unable to find a match: nvidia-imex-550-550.144.03-1
```

tariq1890:
Hey @shivakunv , thanks for this PR. Can we edit the PR title and commit message? We are essentially pinning to an older qemu version which is v6.2.0. I am okay with this change, as it fixes this problem.

tariq1890:
It looks like the `latest` tag was updated `2 days ago` (at the time of writing this comment). Before this update, the `latest` tag hadn't been touched in 2 years. So this explains why our docker multi-arch builds were suddenly failing and why this PR was needed

Reference: https://hub.docker.com/r/tonistiigi/binfmt/tags

cdesiniotis (approved):

PR #229, opened 2025-02-14
[rhel8][arm64] fix failing installs of nvidia-imex packages

tariq1890:


rhel8/Dockerfile
```diff
@@ -82,6 +82,7 @@ RUN if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$TARGETARCH" != "arm64" ]; then \
     dnf install -y ${fmPackage} ${nscqPackage}; fi
 
 RUN if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$DRIVER_BRANCH" -ge "550" ]; then \
+    dnf module enable -y nvidia-driver:${DRIVER_BRANCH}-dkms && \
```
cdesiniotis, line 85: Question -- should we not do this once when setting up the CUDA repository? We are duplicating this line here and before we install FM / nscq.

tariq1890: Yeah, I think that makes sense

tariq1890: Please check now

cdesiniotis (approved):

PR #230, opened 2025-02-14
dependabot update for nested dir

shivakunv:
directories:
        - "*"  

Dependabot does not support wildcard * or **/ paths in the directory field. 
docker section will fail silently

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,7 +20,7 @@ updates:
 
   - package-ecosystem: "docker"
     directories:
-      - "*"
+      - "**/*"
     open-pull-requests-limit: 15
     schedule:
       interval: "daily"


scripts/generatedepndabot.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
ArangoGutierrez, line 1: Is this script needed? how often are we going to update `dependabot.yml`?

ArangoGutierrez (commented):
not a blocking comment

scripts/generatedepndabot.sh
shivakunv: rarely, whenever a new OS or OS version is added, or a new directory with a Dockerfile is created, 
we need to run it locally to generate the dependabot.yaml file and commit the changes.

shivakunv: Removed the generation script and verifying **/*.
if it works, we will proceed with this solution 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files:~:text=To%20specify%20manifests%20in%20the%20current%20directory%20and%20recursive%20subdirectories

ArangoGutierrez (approved):
as recommended at https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files

PR #237, opened 2025-02-18
ubuntu22.04 6.8 kernel support

shivakunv:
### **Features and Code Changes:**

 1. LTS Kernel Matrix: Added support for 5.15 and 6.8 for (ubuntu22.04)
 2. Base Target and LTS KERNEL Version Matrix: Added support for Jammy and Noble.
 3. Hardcoded Values Removed: 
  _base_target_ubuntu22_04, base_target_ubuntu24_04, and lts_kernel_ubuntu24_04 are now dynamically handled instead of being hardcoded._

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -112,6 +112,9 @@ jobs:
         dist:
           - ubuntu22.04
           - ubuntu24.04
+        lts_kernel:
+          - 5.15
+          - 6.8
         ispr:
           - ${{github.event_name == 'pull_request'}}
         exclude:
@@ -124,6 +127,8 @@ jobs:
           - ispr: true
             flavor: oracle
           - driver: 535
+            dist: ubuntu24.04
+          - lts_kernel: 5.15
             dist: ubuntu24.04
     steps:
       - uses: actions/checkout@v4
@@ -164,13 +169,12 @@ jobs:
         env:
           IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/driver
           VERSION: ${COMMIT_SHORT_SHA}
+          LTS_KERNEL: ${{ matrix.lts_kernel }}
         run: |
           if [[ "${{ matrix.dist }}" == "ubuntu22.04" ]]; then
             BASE_TARGET="jammy"
-            LTS_KERNEL="5.15"
           elif [[ "${{ matrix.dist }}" == "ubuntu24.04" ]]; then
             BASE_TARGET="noble"
-            LTS_KERNEL="6.8"
           fi
           make DRIVER_BRANCH=${{ matrix.driver }} KERNEL_FLAVOR=${{ matrix.flavor }} LTS_KERNEL=${LTS_KERNEL} build-base-${BASE_TARGET}
 
--- a/.github/workflows/precompiled.yaml
+++ b/.github/workflows/precompiled.yaml
@@ -26,10 +26,7 @@ jobs:
       driver_branch: ${{ steps.extract_driver_branch.outputs.driver_branch }}
       kernel_flavors: ${{ steps.extract_driver_branch.outputs.kernel_flavors }}
       dist: ${{ steps.extract_driver_branch.outputs.dist }}
-      base_target_ubuntu22_04: ${{ steps.extract_driver_branch.outputs.base_target_ubuntu22_04 }}
-      lts_kernel_ubuntu22_04: ${{ steps.extract_driver_branch.outputs.lts_kernel_ubuntu22_04 }}
-      base_target_ubuntu24_04: ${{ steps.extract_driver_branch.outputs.base_target_ubuntu24_04 }}
-      lts_kernel_ubuntu24_04: ${{ steps.extract_driver_branch.outputs.lts_kernel_ubuntu24_04 }}
+      lts_kernel: ${{ steps.extract_driver_branch.outputs.lts_kernel }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -51,11 +48,10 @@ jobs:
           dist_json=$(printf '%s\n' "${DIST[@]}" | jq -R . | jq -cs .)
           echo "dist=$dist_json" >> $GITHUB_OUTPUT
 
-          # BASE_TARGET and LTS_KERNEL setup
-          echo "base_target_ubuntu22_04=jammy" >> $GITHUB_OUTPUT
-          echo "base_target_ubuntu24_04=noble" >> $GITHUB_OUTPUT
-          echo "lts_kernel_ubuntu22_04=5.15" >> $GITHUB_OUTPUT
-          echo "lts_kernel_ubuntu24_04=6.8" >> $GITHUB_OUTPUT
+          # LTS_KERNEL setup
+          LTS_KERNEL=("5.15" "6.8")
+          lts_kernel_json=$(printf '%s\n' "${LTS_KERNEL[@]}" | jq -R . | jq -cs .)
+          echo "lts_kernel=$lts_kernel_json" >> $GITHUB_OUTPUT
 
   precompiled-build-image:
     needs: set-driver-version-matrix
@@ -65,16 +61,17 @@ jobs:
         driver_branch: ${{ fromJson(needs.set-driver-version-matrix.outputs.driver_branch) }}
         flavor: ${{ fromJson(needs.set-driver-version-matrix.outputs.kernel_flavors) }}
         dist: ${{ fromJson(needs.set-driver-version-matrix.outputs.dist) }}
+        lts_kernel: ${{ fromJson(needs.set-driver-version-matrix.outputs.lts_kernel) }}
         exclude:
           - dist: ubuntu24.04
             driver_branch: 535
+          - lts_kernel: 5.15
+            dist: ubuntu24.04
     steps:
       - uses: actions/checkout@v4
         name: Check out code
       - name: Calculate build vars
         id: vars
-        env:
-          DIST: ${{ matrix.dist }}
         run: |
           echo "LOWERCASE_REPO_OWNER=$(echo "${GITHUB_REPOSITORY_OWNER}" | awk '{print tolower($0)}')" >> $GITHUB_ENV
           REPO_FULL_NAME="${{ github.repository }}"
@@ -83,11 +80,6 @@ jobs:
           GENERATE_ARTIFACTS="false"
           echo "PUSH_ON_BUILD=${GENERATE_ARTIFACTS}" >> $GITHUB_ENV
           echo "BUILD_MULTI_ARCH_IMAGES=${GENERATE_ARTIFACTS}" >> $GITHUB_ENV
-
-          VAR_NAME_BASE_TARGET=$(echo base_target_${DIST} | sed 's/\./_/g')
-          echo "VAR_NAME_BASE_TARGET=${VAR_NAME_BASE_TARGET}" >> $GITHUB_ENV
-          VAR_NAME_LTS_KERNEL=$(echo lts_kernel_${DIST} | sed 's/\./_/g')
-          echo "VAR_NAME_LTS_KERNEL=${VAR_NAME_LTS_KERNEL}" >> $GITHUB_ENV
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
@@ -104,9 +96,13 @@ jobs:
       - name: Build base image and get kernel version
         env:
           IMAGE_NAME: ghcr.io/nvidia/driver
-          BASE_TARGET: ${{ needs.set-driver-version-matrix.outputs[env.VAR_NAME_BASE_TARGET] }}
-          LTS_KERNEL: ${{ needs.set-driver-version-matrix.outputs[env.VAR_NAME_LTS_KERNEL] }}
+          LTS_KERNEL: ${{ matrix.lts_kernel }}
         run: |
+          if [[ "${{ matrix.dist }}" == "ubuntu22.04" ]]; then
+            BASE_TARGET="jammy"
+          elif [[ "${{ matrix.dist }}" == "ubuntu24.04" ]]; then
+            BASE_TARGET="noble"
+          fi
           make DRIVER_BRANCH=${{ matrix.driver_branch }} KERNEL_FLAVOR=${{ matrix.flavor }} LTS_KERNEL=${LTS_KERNEL} build-base-${BASE_TARGET}
 
           trap "docker rm -f base-${BASE_TARGET}-${{ matrix.flavor }}" EXIT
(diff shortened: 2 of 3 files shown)


.github/workflows/image.yaml
```diff
@@ -21,12 +21,14 @@ on:
       - opened
       - synchronize
     branches:
-      - main
-      - release-*
+      # - main
+      # - release-*
```
shivakunv, line 25: test code

tests/scripts/findkernelversion.sh
```diff
@@ -17,30 +18,20 @@ chmod a+x bin/regctl
 export PATH=$(pwd)/bin:${PATH}
 
 # calculate kernel version of latest image
-regctl image get-file ghcr.io/nvidia/driver:base-${BASE_TARGET}-${KERNEL_FLAVOR}-${DRIVER_BRANCH} /var/kernel_version.txt ./kernel_version.txt 2>/dev/null || true
```
shivakunv, line 20: Removed downloading of kernel_version.txt from the Docker base (regctl download in findkernelversion.sh).
Now, artifacts-based download is considered instead.
Reason: When building the base via make build-base-jammy, it now includes two LTS kernels (Ubuntu 22.04: 5.15 and 6.8). Running a parallel matrix will cause make build-base-jammy to overwrite( 5.15 and 6.8) kernel-version.txt, leading to incorrect values in matrix_values files.
Solution: Either modify the Makefile to introduce a new command (make build-base-jammy-lts_kernel) or directly download kernel_version.txt from artifacts.
Since kernel_version... files are already being uploaded as artifacts, we now rely solely on those values._

tests/scripts/findkernelversion.sh
```diff
@@ -17,30 +18,20 @@ chmod a+x bin/regctl
...
-            rm -f $artifact.tar
-            export $(grep -oP 'KERNEL_VERSION=[^ ]+' ./kernel_version.txt)
-            rm -f kernel_version.txt
-            break
-        fi
-    done
-fi
+prefix="kernel-version-${DRIVER_BRANCH}-${LTS_KERNEL}"
+suffix="${kernel_flavor}-${DIST}"
+artifacts=$(gh api -X GET /repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/artifacts --paginate --jq '.artifacts[].name')
```
shivakunv, line 23: paginate option in gh Get API for artifact listing to support more than 30 artifacts.

.github/workflows/image.yaml
```diff
@@ -42,6 +42,12 @@ jobs:
           - ubuntu22.04
           - ubuntu24.04
           - rhel8
+        base_target:
+          - jammy
+          - noble
+        lts_kernel:
+          - 5.15
+          - 6.8
```
cdesiniotis, line 50: I don't believe this is desired. The `image` job does not build precompiled images.

.github/workflows/image.yaml
```diff
@@ -53,6 +59,14 @@ jobs:
...
           - dist: ubuntu24.04
             driver: 535.230.02
+          - base_target: noble
+            dist: ubuntu20.04
+          - base_target: noble
+            dist: ubuntu22.04
+          - base_target: jammy
+            dist: ubuntu24.04
+          - base_target: noble
+            lts_kernel: 5.15
```
cdesiniotis, line 69: Likewise, this should be removed.

.github/workflows/image.yaml
```diff
@@ -93,7 +107,7 @@ jobs:
           IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/driver
           VERSION: ${COMMIT_SHORT_SHA}
         run: |
-          DRIVER_VERSIONS=${{ matrix.driver }} make build-${{ matrix.dist }}-${{ matrix.driver }}
+          BASE_TARGET=${{ matrix.base_target }} LTS_KERNEL=${{ matrix.lts_kernel }} DRIVER_VERSIONS=${{ matrix.driver }} make build-${{ matrix.dist }}-${{ matrix.driver }}
```
cdesiniotis, line 110: Likewise, this should be removed.

.github/workflows/image.yaml
```diff
@@ -111,6 +125,12 @@ jobs:
         dist:
           - ubuntu22.04
           - ubuntu24.04
+        base_target:
+          - jammy
+          - noble
```
cdesiniotis, line 130: IMO defining both `base_target` and `dist` is redundant since there is a one-to-one mapping between the two of them. If we really need to define both strings, then one of them should be assigned in a Makefile or script.

.github/workflows/precompiled.yaml
```diff
@@ -199,21 +199,21 @@ jobs:
...
+          KERNEL_VERSIONS=($(get_kernel_versions_to_test $BASE_TARGET KERNEL_FLAVORS[@] DRIVER_BRANCHES[@] $DIST $LTS_KERNEL))
           if [ -z "$KERNEL_VERSIONS" ]; then
             # no new kernel release
             echo "Skipping e2e tests"
             exit 0
           fi
           # Convert array to JSON format and assign
-          echo "[]" > ./matrix_values_$DIST.json
-          printf '%s\n' "${KERNEL_VERSIONS[@]}" | jq -R . | jq -s . > ./matrix_values_$DIST.json
+          echo "[]" > ./matrix_values_${{ matrix.dist }}_${{ matrix.base_target }}_${{ matrix.lts_kernel }}.json
```
cdesiniotis, line 209: As dist and base_target are redundant, can we just include one of them in the file name?

.github/workflows/image.yaml
shivakunv: removed

base/generate-ci-config
```diff
@@ -16,7 +16,7 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update -y -qq > /dev/null
 
 # Generate a list of all kernel versions which have NVIDIA precompiled driver packages available.
-SUPPORTED_KERNELS=$(apt-cache search linux-objects-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR} | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
+SUPPORTED_KERNELS=$(apt-cache search  "linux-image-${LTS_KERNEL}.*-${KERNEL_FLAVOR}" | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
```
shivakunv, line 19: rely on linux-image- for supported kernel version.

Sometimes, the Linux object package is released, but the corresponding Linux image is only available for the low-latency variant.

e.g:- in the current situation, linux-object-6.8.0-55-generic is available, but linux-image-6.8.0-55-generic is not. 
Instead, linux-image-6.8.0-52-generic and linux-image-6.8.0-55-lowlatency are available.

base/generate-ci-config
```diff
@@ -16,7 +16,7 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update -y -qq > /dev/null
 
 # Generate a list of all kernel versions which have NVIDIA precompiled driver packages available.
-SUPPORTED_KERNELS=$(apt-cache search linux-objects-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR} | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
+SUPPORTED_KERNELS=$(apt-cache search linux-image-${LTS_KERNEL}.*-${KERNEL_FLAVOR} | awk '{print $1}' | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
```
shivakunv, line 19: rely on linux-image- for supported kernel version.

Sometimes, the Linux object package is released, but the corresponding Linux image is only available for the low-latency variant.

e.g:- in the current situation for ubuntu22.04, linux-object-6.8.0-55-generic is available, but linux-image-6.8.0-55-generic is not.
Instead, linux-image-6.8.0-52-generic and linux-image-6.8.0-55-lowlatency are available.

.github/workflows/precompiled.yaml
```diff
@@ -104,9 +96,13 @@ jobs:
...
           IMAGE_NAME: ghcr.io/nvidia/driver
-          BASE_TARGET: ${{ needs.set-driver-version-matrix.outputs[env.VAR_NAME_BASE_TARGET] }}
-          LTS_KERNEL: ${{ needs.set-driver-version-matrix.outputs[env.VAR_NAME_LTS_KERNEL] }}
+          LTS_KERNEL: ${{ matrix.lts_kernel }}
         run: |
+          if [[ "${{ matrix.dist }}" == "ubuntu22.04" ]]; then
+            BASE_TARGET="jammy"
+          elif [[ "${{ matrix.dist }}" == "ubuntu24.04" ]]; then
+            BASE_TARGET="noble"
+          fi
```
cdesiniotis, line 105: Out of scope for this PR: we should clean this up so that we do not need to specify both BASE_TARGET and DIST. We should just use one. This may require some refactoring of our `build-base-*` make targets.

base/generate-ci-config
cdesiniotis: The purpose of this code is to generate the list of precompiled NVIDIA driver packages. There is typically a delay between when a new kernel (aka `linux-image` package) is released and when the NVIDIA driver packages for that kernel are made available. Therefore, we need to revert this change. 

FWIW, I do not see any kernel available for 6.8.0-55-generic on Ubuntu 22.04:
```
$ sudo apt-cache search linux-objects-nvidia-550-server | grep 6.8.0-55
$ sudo apt-cache search linux-image | grep 6.8.0-55
```
The latest kernel version available that I see is `6.8.0-52-generic`

cdesiniotis: I have reverted the change.

cdesiniotis: If this logic can be improved in some way, let's discuss and aim to do it in a follow-up.

shivakunv: @cdesiniotis : 
local system reports 6.8.0-52-generic , but docker shell ( int shell) reports :- 

export DRIVER_BRANCH=535
export LTS_KERNEL=6.8
export KERNEL_FLAVOR=generic
apt-cache search linux-objects-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR} | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2 

6.8.0-38-generic
6.8.0-39-generic
6.8.0-40-generic
6.8.0-45-generic
6.8.0-47-generic
6.8.0-48-generic
6.8.0-49-generic
6.8.0-50-generic
6.8.0-51-generic
6.8.0-52-generic
6.8.0-55-generic

due to this, the e2e aws system is unable to find 6.8.0-55-generic (as it has not been released yet). 
failed pipeline: 
https://github.com/NVIDIA/gpu-driver-container/actions/runs/13885688001/job/38850157804

.github/workflows/precompiled.yaml
shivakunv: okay , TODO added

base/generate-ci-config
```diff
@@ -16,14 +16,17 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update -y -qq > /dev/null
 
 # Generate a list of all kernel versions which have NVIDIA precompiled driver packages available.
-SUPPORTED_KERNELS=$(apt-cache search linux-objects-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR} | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
+SUPPORTED_KERNELS_LINUX_OBJECTS_LIST=$(apt-cache search linux-objects-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR} | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
```
tariq1890, line 19: linux-objects-nvidia is a package that is only installed when installing the proprietary modules. Given than OpenRM is the officially preferred kernel module. I'd propose checking for either of the two packages instead

- `linux-signatures-nvidia.*${LTS_KERNEL}.*${KERNEL_FLAVOR}` (preferred as it's installed in both OpenRM and proprietary)
- `linux-modules-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR}`

.github/workflows/image.yaml
```diff
@@ -111,6 +111,9 @@ jobs:
         dist:
           - ubuntu22.04
           - ubuntu24.04
+        lts_kernel:
+          - 5.15
+          - 6.8
```
shivakunv, line 116: add 6.8 support in .nvidia.yaml

.github/workflows/image.yaml
```diff
@@ -166,11 +171,10 @@ jobs:
         run: |
           if [[ "${{ matrix.dist }}" == "ubuntu22.04" ]]; then
             BASE_TARGET="jammy"
-            LTS_KERNEL="5.15"
           elif [[ "${{ matrix.dist }}" == "ubuntu24.04" ]]; then
             BASE_TARGET="noble"
-            LTS_KERNEL="6.8"
           fi
+          LTS_KERNEL=${{ matrix.lts_kernel }}
```
cdesiniotis, line 177: nit: can we move this assignment statement in the `env` stanza instead?

base/generate-ci-config
```diff
@@ -16,14 +16,17 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update -y -qq > /dev/null
 
 # Generate a list of all kernel versions which have NVIDIA precompiled driver packages available.
-SUPPORTED_KERNELS=$(apt-cache search linux-objects-nvidia-${DRIVER_BRANCH}-server.*${LTS_KERNEL}.*${KERNEL_FLAVOR} | awk '{print $1}' | grep -v "open" | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2)
+SUPPORTED_KERNELS_LINUX_SIGNATURES_LIST=$(apt-cache search "linux-signatures-nvidia.*${LTS_KERNEL}.*${KERNEL_FLAVOR}" | awk '{print $1}'  | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2| grep "${KERNEL_FLAVOR}$")
```
cdesiniotis, line 19: Question: what is the rationale for using the `linux-signatures-nvidia` package instead of the `linux-objects-nvidia` package?

tariq1890: I had suggested this earlier as the `linux-objects-nvidia` package is only specific to the proprietary modules whereas `linux-signatures-nvidia` is used in both. 

I didn't realise that we were looking up `linux-objects-nvidia`, so one could argue that this particular change can be made in a follow-up PR

cdesiniotis: Ah okay. I think I am okay with using the `linux-signatures-nvidia` package. Was just curious what led to this change.

tariq1890 (approved):

PR #246, opened 2025-02-28
[ubuntu22.04] add support for nvlink5+ systems

tariq1890:


ubuntu22.04/Dockerfile
```diff
@@ -85,11 +85,14 @@ RUN if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$TARGETARCH" != "arm64" ]; then \
...
     libnvidia-nscq-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 && \
+    if [ "$DRIVER_BRANCH" -ge "570" ]; then \
+    apt-get install -y --no-install-recommends libnvsdm-${DRIVER_BRANCH}=${DRIVER_VERSION}-1; \
+    fi && \
     rm -rf /var/lib/apt/lists/*; fi
 
 RUN if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$DRIVER_BRANCH" -ge "550" ]; then \
     apt-get update && \
-    apt-get install -y --no-install-recommends nvidia-imex-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 && \
+    apt-get install -y --no-install-recommends nvidia-imex-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 nvlsm && \
```
cdesiniotis, line 95: Why is the installation of nvlsm and libnvsdm separated?

ubuntu22.04/nvidia-driver
```diff
@@ -321,7 +351,18 @@ _load_driver() {
         _start_vgpu_topology_daemon
     fi
 
-    if _assert_nvswitch_system; then
+    if _assert_nvlink5_system; then
+        _ensure_nvlink5_prerequisites || return 1
```
cdesiniotis, line 355: As mentioned yesterday, instead of returning an error here should we consider waiting until the drivers are loaded?

ubuntu22.04/nvidia-driver
```diff
@@ -195,6 +195,36 @@ _assert_nvswitch_system() {
...
+                return 0
+            fi
+        fi
+    done
+    return 1
+)
+
+_ensure_nvlink5_prerequisites() (
+    echo "Ensure that the ib_umad kernel module is loaded"
+    required_module_name="ib_umad"
```
cdesiniotis, line 217: Do we not need to check for the mofed module too?

ubuntu22.04/Dockerfile
tariq1890: I believe gb200 needs nvlsm. So it can't be skipped for arm unlike libnvsdm

ubuntu22.04/nvidia-driver
tariq1890: Sure, we could. What I have observed was that ib_umad was only loaded after MOFED was installed and the node rebooted

tariq1890: So I figured the mlx5_core module check was redundant

cdesiniotis: If ib_umad depends on mlx5_core, then I guess only checking for ib_umad would suffice.

ubuntu22.04/Dockerfile
cdesiniotis: Since FM is supported on arm, maybe its time to start installing FM in our ARM based containers? That would simplify the changes here and allow us to install all of these additional packages in the same conditional block.

tariq1890: Sure, that sounds good to me

ubuntu22.04/nvidia-driver
tariq1890: Done, please re-review

ubuntu22.04/nvidia-driver
tariq1890: Checking for both modules now. Just for good measure

ubuntu22.04/Dockerfile
```diff
@@ -80,16 +80,22 @@ RUN if [ "$DRIVER_TYPE" != "vgpu" ]; then \
...
+RUN if [ "$DRIVER_TYPE" != "vgpu" ]; then \
     apt-get update && \
     apt-get install -y --no-install-recommends nvidia-fabricmanager-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 \
     libnvidia-nscq-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 && \
     rm -rf /var/lib/apt/lists/*; fi
 
 RUN if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$DRIVER_BRANCH" -ge "550" ]; then \
     apt-get update && \
-    apt-get install -y --no-install-recommends nvidia-imex-${DRIVER_BRANCH}=${DRIVER_VERSION}-1 && \
+    apt-get install -y --no-install-recommends nvlsm \
```
cdesiniotis, line 91: Question: is there a specific version of nvlsm we need to install? Is it tightly coupled to the driver version? Or is it desired to pick up the latest?

ubuntu22.04/Dockerfile
```diff
@@ -80,16 +80,22 @@ RUN if [ "$DRIVER_TYPE" != "vgpu" ]; then \
     cd drivers && \
     /tmp/install.sh download_installer; fi
 
-# Fabric manager packages are not available for arm64
-RUN if [ "$DRIVER_TYPE" != "vgpu" ] && [ "$TARGETARCH" != "arm64" ]; then \
+RUN if [ "$DRIVER_TYPE" != "vgpu" ]; then \
```
cdesiniotis, line 83: nit: not a blocker, but I would have preferred if this change was a separate commit.

ubuntu22.04/nvidia-driver
```diff
@@ -195,6 +195,34 @@ _assert_nvswitch_system() {
...
+
+_ensure_nvlink5_prerequisites() (
+    until  lsmod | grep mlx5_core > /dev/null 2>&1 && lsmod | grep ib_umad > /dev/null 2>&1;
+    do
+        echo "waiting for the mlx5_core and ib_umad kernel modules to be loaded"
+        sleep 10
+    done
+
+    echo "Installing the infiniband-diags package"
+    apt-get -qq install --no-install-recommends infiniband-diags > /dev/null
```
cdesiniotis, line 223: Why is this package required?

ubuntu22.04/Dockerfile
tariq1890: nvlsm doesn't seem to be pinned to any particular when you look it up [here](https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/)

ubuntu22.04/nvidia-driver
tariq1890: the `fabricmanager` runs the `ibstat` command to fetch the guids of the NVLink5 switches. ibstat is a part of the `infiniband-diags` pakage

ubuntu22.04/Dockerfile
tariq1890: fair enough. I can raise a separate PR for that

tariq1890: Done, please check - https://github.com/NVIDIA/gpu-driver-container/pull/278

ubuntu22.04/nvidia-driver
cdesiniotis: If `infiniband-diags` is small in size, should we consider installing it at build time?

tariq1890: Sure, it seems small enough

tariq1890: Done

cdesiniotis (approved):

tariq1890:
Tested this container image on both NvLink5 and non-nvlink systems. Working as expected.

PR #249, opened 2025-03-05
Precompile kernel version fix for multiple base target

shivakunv:
_Removed downloading of kernel_version.txt from the Docker base.
Now, artifacts-based download is considered instead.
Reason: When building the base via make build-base-jammy, it now includes two LTS kernels (Ubuntu 22.04: 5.15 and 6.8). Running a parallel matrix will cause make build-base-jammy to overwrite( 5.15 and 6.8) kernel-version.txt, leading to incorrect values in matrix_values files during download with regctl in determine-e2e-test-matrix job. with regctl command we are downloading kernel_version.txt file from docker base. 
Solution: Either modify the Makefile to introduce a new command (make build-base-${base_target}-${lts_kernel}) or directly download kernel_version.txt from artifacts.
Since kernel_version... files are already being uploaded as artifacts, we now rely on those values._

### DIFF
--- a/.github/workflows/precompiled.yaml
+++ b/.github/workflows/precompiled.yaml
@@ -199,7 +199,7 @@ jobs:
           done))
           fi
           source ./tests/scripts/ci-precompiled-helpers.sh
-          KERNEL_VERSIONS=($(get_kernel_versions_to_test $BASE_TARGET KERNEL_FLAVORS[@] DRIVER_BRANCHES[@] $DIST))
+          KERNEL_VERSIONS=($(get_kernel_versions_to_test KERNEL_FLAVORS[@] DRIVER_BRANCHES[@] $DIST $LTS_KERNEL))
           if [ -z "$KERNEL_VERSIONS" ]; then
             # no new kernel release
             echo "Skipping e2e tests"
--- a/tests/scripts/findkernelversion.sh
+++ b/tests/scripts/findkernelversion.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 
 if [[ $# -ne 4 ]]; then
-	echo " BASE_TARGET KERNEL_FLAVOR DRIVER_BRANCH DIST are required"
+	echo " KERNEL_FLAVOR DRIVER_BRANCH DIST LTS_KERNEL are required"
 	exit 1
 fi
 
-export BASE_TARGET="${1}"
-export KERNEL_FLAVOR="${2}"
-export DRIVER_BRANCH="${3}"
-export DIST="${4}"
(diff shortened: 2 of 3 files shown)


tests/scripts/findkernelversion.sh
```diff
@@ -17,30 +18,20 @@ chmod a+x bin/regctl
 export PATH=$(pwd)/bin:${PATH}
 
 # calculate kernel version of latest image
-regctl image get-file ghcr.io/nvidia/driver:base-${BASE_TARGET}-${KERNEL_FLAVOR}-${DRIVER_BRANCH} /var/kernel_version.txt ./kernel_version.txt 2>/dev/null || true
```
shivakunv, line 20: removed:
driver:base-${BASE_TARGET}-${KERNEL_FLAVOR}-${DRIVER_BRANCH}  
can be : 
driver:base-jammy-nvidia-535 :  true for both 6.8 and 5.15  . 

Since kernel_version... files are already being uploaded as artifacts, we now rely on uploaded artifacts kernel version

.github/workflows/precompiled.yaml
```diff
@@ -199,21 +199,21 @@ jobs:
...
+          KERNEL_VERSIONS=($(get_kernel_versions_to_test $BASE_TARGET KERNEL_FLAVORS[@] DRIVER_BRANCHES[@] $DIST $LTS_KERNEL))
           if [ -z "$KERNEL_VERSIONS" ]; then
             # no new kernel release
             echo "Skipping e2e tests"
             exit 0
           fi
           # Convert array to JSON format and assign
-          echo "[]" > ./matrix_values_$DIST.json
-          printf '%s\n' "${KERNEL_VERSIONS[@]}" | jq -R . | jq -s . > ./matrix_values_$DIST.json
+          echo "[]" > ./matrix_values_${{ matrix.dist }}_${{ needs.set-driver-version-matrix.outputs[env.VAR_NAME_BASE_TARGET] }}_$LTS_KERNEL.json
```
shivakunv, line 209: currently , we have hardcoded values for "base_target_ubuntu22_04": "jammy", "base_target_ubuntu24_04": "noble"  and "lts_kernel_ubuntu22_04": "5.15", "lts_kernel_ubuntu24_04": "6.8"


dist will be removed and base target will be mapped to dist in PR:- 
https://github.com/NVIDIA/gpu-driver-container/pull/237

@cdesiniotis 
https://github.com/NVIDIA/gpu-driver-container/pull/237#discussion_r1994739779

cdesiniotis (approved):

PR #256, opened 2025-03-10
[ubuntu24.04][precompiled] rm /etc/apt/sources.list.d/remove ubuntu.sources

tariq1890:


cdesiniotis:
I am assuming this change is also present in ubuntu22.04?

tariq1890:
No, this file is only present in the ubuntu 24.04  cuda base image

ubuntu24.04/precompiled/Dockerfile
```diff
@@ -55,12 +55,11 @@ ADD local-repo.sh /tmp
 RUN mkdir -p /usr/local/repos && \
     /tmp/local-repo.sh download_driver_package_deps && \
     /tmp/local-repo.sh build_local_apt_repo && \
+    # Remove the ubuntu public ubuntu apt repo to ensure we only pull from the local apt repo
+    rm /etc/apt/sources.list.d/ubuntu.sources && \
```
cdesiniotis, line 59: Wouldn't it be better to remove all filed under `/etc/apt/sources.list.d` since we configure the local repo in the top level `/etc/apt/sources.list` file?

tariq1890: Yeah, I can do that. Right now, ubuntu.sources is the only file under that directory. But, that could change in the future

cdesiniotis (approved):

PR #260, opened 2025-03-14
[ci] fix internal precompiled jobs

cdesiniotis:
If 'rules' are defined for a Gitlab CI job, then the job will only be added to the pipeline if one of the rules matches. This commit ensures that for scheduled pipelines, at least one rule matches for the precompiled jobs which are meant to run on the schedule.


shivakunv (approved):
LGTM

tariq1890 (approved):

PR #279, opened 2025-03-14
[precompiled] add support for R570 drivers

tariq1890:


tariq1890:
No hits for the `nvidia-utils-570-server` package in the build-base stage, but the same shows up on [packages.ubuntu.com](https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=nvidia-utils-570&searchon=names)

tariq1890:
update: Canonical has redirected the `nvidia-driver-570-server` packages to the `restricted` repo (which is the expected configuration) and the build jobs have started to pass

tariq1890:
Tested on Ubuntu 22.04 and Ubuntu 24.04 nodes

.github/workflows/image.yaml
```diff
@@ -102,6 +102,7 @@ jobs:
         driver: 
           - 535
           - 550
+          - 570
```
cdesiniotis, line 105: Question: do we also need to update the version matrix in `precompiled.yaml` (our scheduled pipelines)?

tariq1890: Ah yes, thanks for catching that. I will update that as well

cdesiniotis (approved):

PR #288, opened 2025-03-25
[vgpu-manager/ubuntu22.04] Install gcc-12 in vgpu driver manager container

gfaugere:
This was fixed in dd69782dc6a21aa92ded68fb9db58bd4b1a23a4a (cc @tariq1890) for regular installs, but vgpu-manager still fails to compile drivers in Ubuntu22.04 with Kernels in 5.19.x/6.5.x

This fix seemed appropriate and straight-forward enough to simply port it to the vgpu-manager image


cdesiniotis (approved):
@gfaugere Thank you for the contribution!

cdesiniotis:
@gfaugere I see that your commit is unverified due to: `This user has not yet uploaded their public signing key.` Would you be able to upload your public signing key?

PR #290, opened 2025-04-01
fix bug in download_apt_with_dep method

tariq1890:
This PR fixes a bug in the method when we want to download the dependent packages of a package when a version of the latter is specified in the parameter list

Without this fix, the method ends up downloading the latest stable release of the input package along with its dependencies. This is undesirable as we only want to fetch the dependent packages when running the second apt-get download command (L18) of the method

We can now fetch the dependent packages of nvidia-fabricmanager and libnvidia-nscq (pinning to a specific version) packages


cdesiniotis (approved):
Thanks for the detailed description!

PR #298, opened 2025-04-10
print the nvidia-persisentenced stop message only if pid is detected

tariq1890:
This is consistent with the rest of the daemons (gridd, fabric manager, nvlsm) managed by the driver container


ubuntu20.04/precompiled/nvidia-driver
```diff
@@ -349,4 +349,4 @@ if [ $? -ne 0 ]; then
     usage
 fi
 
-$command
\ No newline at end of file
+$command
```
tariq1890, line 352: GoLand auto-adds this

cdesiniotis (approved):
Thanks!

PR #345, opened 2025-06-11
Addition of 64k kernel-flavor in KERNEL_VERSION

shivakunv:
The driver container fails to install on kernels of the generic-64k flavor. The root cause is that the [resolve_kernel_version() function](https://github.com/NVIDIA/gpu-driver-container/blob/e9db94520c008aec0bc6279bfbade65600c8135f/ubuntu22.04/nvidia-driver#L47-L61) is stripping the "-64k" suffix from the kernel version string. So when we attempt to install the kernel-header packages, we are installing the wrong version of the packages.

script tested on arm64 target : 

```
./run.sh 6.8.0-51-generic-64k

run.sh 

#!/bin/bash
get_kernel() {
        local version=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null | \
	sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)
	local flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//' | grep -Ev "^generic|virtual")
	local kernel_flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//')
	kernel_flavor="${kernel_flavor//virtual/generic}"

	echo "Resolving Linux kernel version..."
	if [ -z "${version}" ]; then
		echo "Could not resolve Linux kernel version" >&2
		return 1
	fi

	KERNEL_VERSION="${version}-${flavor:-$kernel_flavor}"
	echo "Proceeding with Linux kernel version ${KERNEL_VERSION}"
}

export KERNEL_VERSION=$1
get_kernel
```

output : 
_Resolving Linux kernel version...
Proceeding with Linux kernel version 6.8.0-51-generic-64k_

### DIFF
--- a/ubuntu20.04/nvidia-driver
+++ b/ubuntu20.04/nvidia-driver
@@ -47,7 +47,8 @@ _update_ca_certificates() {
 _resolve_kernel_version() {
     local version=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null | \
       sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)
-    local flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//' | grep -Ev "^generic|virtual")
+    local kernel_flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//')
+    kernel_flavor="${kernel_flavor//virtual/generic}"
 
     echo "Resolving Linux kernel version..."
     if [ -z "${version}" ]; then
@@ -55,7 +56,7 @@ _resolve_kernel_version() {
         return 1
     fi
 
-    KERNEL_VERSION="${version}-${flavor:-generic}"
+    KERNEL_VERSION="${version}-${kernel_flavor}"
     echo "Proceeding with Linux kernel version ${KERNEL_VERSION}"
     return 0
 }
--- a/ubuntu16.04/nvidia-driver
+++ b/ubuntu16.04/nvidia-driver
@@ -28,14 +28,16 @@ _cleanup_package_cache() {
 _resolve_kernel_version() {
     local version=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null | \
       sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)
-    local flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//' | grep -Ev "^generic|virtual")
+    local kernel_flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//')
+    kernel_flavor="${kernel_flavor//virtual/generic}"
 
     echo "Resolving Linux kernel version..."
     if [ -z "${version}" ]; then
         echo "Could not resolve Linux kernel version" >&2
         return 1
     fi
-    KERNEL_VERSION="${version}-${flavor:-generic}"
+
+    KERNEL_VERSION="${version}-${kernel_flavor}"
     echo "Proceeding with Linux kernel version ${KERNEL_VERSION}"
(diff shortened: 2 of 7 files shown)


shivakunv:
@cdesiniotis 

The changes seem fine to me.

It is possible for _uname -r_ and _apt-cache show_ to refer to different base kernel versions.
This is expected behavior , the version field can differ from what uname -r is showing (if the booted kernel is removed/stale/custom/headers) 

Now, coming to the second part: I am not able to test it, as every time I run apt install linux-image-virtual, uname -r still shows the -generic version ( both arm and amd)

However, I did notice in the logs that linux-image-virtual depends on both the generic kernel and generic headers. 
```
sudo apt install linux-image-virtual
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  linux-image-6.8.0-60-generic linux-modules-6.8.0-60-generic
Suggested packages:
  fdutils **linux-tools linux-headers-6.8.0-60-generic** linux-modules-extra-6.8.0-60-generic
The following NEW packages will be installed:
  linux-image-6.8.0-60-generic linux-image-virtual linux-modules-6.8.0-60-generic
0 upgraded, 3 newly installed, 0 to remove and 0 not upgraded.
Need to get 105 MB of archives.
After this operation, 105 MB of additional disk space will be used.
```

So, I simplified the previous code that replaced _virtual_ with _generic_ to  handle kernel_flavor suffixes (e.g., generic-64k).

ubuntu20.04/nvidia-driver
cdesiniotis, line 1: Do we also need to update `ubuntu22.04`?

ubuntu20.04/nvidia-driver
```diff
@@ -47,15 +47,16 @@ _update_ca_certificates() {
 _resolve_kernel_version() {
     local version=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null | \
       sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)
-    local flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//' | grep -Ev "^generic|virtual")
+    local kernel_flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//')
+    kernel_flavor="${kernel_flavor//virtual/generic}"
```
cdesiniotis, line 51: So if I understand correctly, before this change we were changing any `generic-*` or `virtual` flavors to just `generic` (this caused issues with the `generic-64k` kernels since the `-64k` suffix was stripped. With your changes, we are only changing `virtual` flavors to `generic` -- which through your tests seems to be needed.

cdesiniotis (commented):
Thanks @shivakunv for the work on this and including details from your testing. These changes lgtm. Do we also need to make this change for ubuntu22.04?

Let's wait for @tariq1890 to review as well.

ubuntu20.04/nvidia-driver
shivakunv: yes .

ubuntu20.04/nvidia-driver
shivakunv: @tariq1890 

shall I remove for all the other changes for the other ubuntu version  ? 

image

tariq1890: They can remain. I don't see any issue with adding those changes in the other Ubuntu versions

tariq1890 (dismissed):

ubuntu20.04/nvidia-driver
cdesiniotis: @shivakunv I still don't see this change applied to the `ubuntu22.04` folder. Can you please add the change there as well?

shivakunv: @cdesiniotis @tariq1890 
vscode sync issue. 
updated the code . PTAL

cdesiniotis (approved):

PR #371, opened 2025-07-03
Slack notification for new kernel release

shivakunv:


.github/workflows/precompiled.yaml
```diff
@@ -269,6 +269,32 @@ jobs:
...
+      - name: Slack notification
+        uses: slackapi/slack-github-action@v2.1.0
+        with:
+          token: ${{ secrets.SLACK_BOT_TOKEN }}
+          method: chat.postMessage
+          payload: |
+            {
+              "channel": "${{ secrets.SLACK_CHANNEL_ID }}",
+              "text": "<!here> New kernel release detected: ${{ matrix.kernel_version }}"
+            }
```
cdesiniotis: Not a strong opinion here -- but if it is not too hard to implement, what about sending one slack message that contains all of the new kernel versions (instead of one message per kernel version)? Also, let's provide a link to this GitHub Action run so that one can refer to it.
```
https://github.com/${{github.repository}}/actions/runs/${{github.run_id}}
```

tariq1890: What if we posted a Slack notification when a new Kernel release is detected and another Slack notification after the precompiled driver image is built (whether the image build was successful or not) ?

tariq1890: I don't think `slack-notification` should be its own stage, but rather a part of the existing stages in our actions workflow

cdesiniotis: > What if we posted a Slack notification when a new Kernel release is detected and another Slack notification after the precompiled driver image is built (whether the image build was successful or not) ?

Yes, this would be helpful to see both the kernel versions AND the image tags that got built for them.

shivakunv: two notifications:-  ??

1. New kernel released ( current implementation) : slack-notify->e2e 
2. slack-notify->e2e->  publish-precompiled-image -> slack-notify 

or send in case of e2e failure 

In 2nd case we have to message individually for every successful build and publish . 

points to remember: 
_In the case of an e2e failure , we are exiting/cancelling the pipeline anyways , so I dont think slack notification will work in that case . For the successful case , we already identify and and send the kernel version . 
So the my question is : Is it necessary to send a slack message during successful publishing , since it will only succeed if our e2e tests pass ?_ 

@cdesiniotis @tariq1890

shivakunv: PTAL @cdesiniotis  @tariq1890  : current change:
  
  1. seprate step added for slack notifications 
  2. Notifications for combined kernel versions in collect-e2e-test-matrix
  3. Notification for every successful publish 
  
  passed pipeline:
     https://github.com/NVIDIA/gpu-driver-container/actions/runs/16072195120

PR #380, opened 2025-07-17
Updated versions to 570.172.08 and 535.261.03

JunAr7112:


.github/workflows/image.yaml
```diff
@@ -34,9 +34,9 @@ jobs:
     strategy:
       matrix:
         driver:
-          - 535.247.01
+          - 535.261.03
```
tariq1890, line 37: You need to update the R535 version in L52 and L57 as well

tariq1890:
To unblock your PR's CI checks, this PR will need to get merged first

.github/workflows/image.yaml
JunAr7112: Alright I updated this. Thanks for pointing out!

versions.mk
```diff
@@ -13,6 +13,6 @@
 # limitations under the License.
 
 # DRIVER_VERSIONS contains latest version in all active datacenter branches
-DRIVER_VERSIONS ?= 535.247.01 550.163.01 570.158.01 575.57.08
+DRIVER_VERSIONS ?= 535.261.03 550.163.01 570.172.08 575.57.08
 
 GOLANG_VERSION := 1.24.4
```
tariq1890: Looks like golang 1.24.5 has been released. Can you update this as well?

JunAr7112: Ok I'll update this

cdesiniotis (approved):

PR #386, opened 2025-07-28
Include lspci in vgpu-manager image

empovit:
In disconnected environments `dnf install` cannot be used without mirroring RPM repositories. As the vGPU manager now requires lspci command:

* Include the RPMs in the container image
* If lspci command not found, install it from the RPMs

### DIFF
--- a/vgpu-manager/rhel8/Dockerfile
+++ b/vgpu-manager/rhel8/Dockerfile
@@ -5,7 +5,11 @@ ENV DRIVER_VERSION=$DRIVER_VERSION
 ARG DRIVER_ARCH=x86_64
 ENV DRIVER_ARCH=$DRIVER_ARCH
 
-RUN mkdir -p /driver
+RUN mkdir -p /driver/rpms/pciutils
+WORKDIR /driver/rpms/pciutils
+
+RUN dnf download --resolve pciutils && dnf clean all
+
 WORKDIR /driver
 COPY NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}-vgpu-kvm.run .
 RUN chmod +x NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}-vgpu-kvm.run
--- a/vgpu-manager/rhel8/ocp_dtk_entrypoint
+++ b/vgpu-manager/rhel8/ocp_dtk_entrypoint
@@ -100,7 +100,7 @@ dtk-build-driver() {
 
     # ensure lspci is installed, as 'sriov-manage' script requires it
     if ! $(lspci >/dev/null); then
-      dnf install -y pciutils && rm -rf /var/cache/yum/*
+      rpm -ivh ${DRIVER_TOOLKIT_SHARED_DIR}/driver/rpms/pciutils/*.rpm
     fi
 
     # upon catching a signal, terminate child process to trigger driver cleanup


SDBrett:
This update uses a UBI8 image to download the `pciutils` rpms and the driver toolkit container image is built on UBI9.

Could this introduce a potential issue when trying to install them on the driver toolkit container?

vgpu-manager/rhel8/Dockerfile
```diff
@@ -5,7 +5,11 @@ ENV DRIVER_VERSION=$DRIVER_VERSION
 ARG DRIVER_ARCH=x86_64
 ENV DRIVER_ARCH=$DRIVER_ARCH
 
-RUN mkdir -p /driver
+RUN mkdir -p /driver/rpms/pciutils
+WORKDIR /driver/rpms/pciutils
+
+RUN dnf download --resolve pciutils && dnf clean all
```
tariq1890, line 11: Why not just install the rpm package instead of downloading it and running it later?

Do we want to use pciutils in the DTK container? If so, we should consider baking pciutils into DTK

SDBrett:
This PR is based on work I did for a client who did not have access to an
RPM mirror to install pciutils.

The preferred fix(es) would be for the NVIDIA operator to allow the toolkit
image to be specified via the ClusterPolicy resource. And / or the operator
to provide the toolkit image with pciutils already installed.

vgpu-manager/rhel8/Dockerfile
cdesiniotis: Yes, currently lspci is needed at runtime by the sriov-manage script that we invoke in the DTK container. Vitaly asked the DTK team if they could include the pciutils package in the DTK image, but they said no.

SDBrett:
> It means the DTK and vgpu-manager images will have to be updated manually during every cluster upgrade. The driver toolkit image has to match the cluster's kernel version, and is automatically managed for each OpenShift version. As the purpose of DTK is building drivers, the team is saying that `lspci` does not belong in the image.
> 
> I like @tariq1890 's idea to bump the base image.
> 
> Other options I can think of:
> 
> 1. Multi-stage build with UBI 9 (not DTK). This option will require re-building vgpu-manager when a cluster (and DTK) goes from RHEL 9 to RHEL 10. Unfortunately, the same issue exists even if the vgpu-manager itself becomes UBI 9.
> 2. Have a tiny image for just `lspci`, or have `lspci` in the vgpu-manager image and pass _the data_ (not the binary) between containers.

The multistage build is how I got things working for the customer.  Below is the diff of those changes.

```
diff --git a/vgpu-manager/rhel8/Dockerfile b/vgpu-manager/rhel8/Dockerfile
index da9c12f..76f68a0 100644
--- a/vgpu-manager/rhel8/Dockerfile
+++ b/vgpu-manager/rhel8/Dockerfile
@@ -1,3 +1,8 @@
+FROM registry.redhat.io/ubi9/ubi:9.6 AS ubi9
+RUN mkdir -p /rpms/pciutils
+WORKDIR /rpms/pciutils
+RUN dnf download --resolve pciutils
+
 FROM nvcr.io/nvidia/cuda:12.9.1-base-ubi8
 
 ARG DRIVER_VERSION
@@ -5,7 +10,9 @@ ENV DRIVER_VERSION=$DRIVER_VERSION
 ARG DRIVER_ARCH=x86_64
 ENV DRIVER_ARCH=$DRIVER_ARCH
 
-RUN mkdir -p /driver
+RUN mkdir -p /driver/rpms
+COPY --from=ubi9 /rpms/ /driver/rpms
+
 WORKDIR /driver
 COPY NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}-vgpu-kvm.run .
 RUN chmod +x NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}-vgpu-kvm.run
diff --git a/vgpu-manager/rhel8/ocp_dtk_entrypoint b/vgpu-manager/rhel8/ocp_dtk_entrypoint
index e5d502f..3a21bd2 100755
--- a/vgpu-manager/rhel8/ocp_dtk_entrypoint
+++ b/vgpu-manager/rhel8/ocp_dtk_entrypoint
@@ -96,6 +96,8 @@ dtk-build-driver() {
        "$DRIVER_TOOLKIT_SHARED_DIR/nvidia-driver" \
        "${DRIVER_TOOLKIT_SHARED_DIR}/bin"
 
+    rpm -ivh ${DRIVER_TOOLKIT_SHARED_DIR}/driver/rpms/pciutils/*.rpm
+
     export PATH="${DRIVER_TOOLKIT_SHARED_DIR}/bin:$PATH";
 
     # ensure lspci is installed, as 'sriov-manage' script requires it

```

PR #417, opened 2025-08-26
[ubuntu24.04] replace cuda base image in favour of upstream ubuntu image

tariq1890:


ubuntu24.04/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM nvcr.io/nvidia/cuda:13.0.0-base-ubuntu24.04 as build
+ARG BASE_IMAGE=ubuntu:noble-20250716
```
cdesiniotis, line 1: Question: will dependabot still be able to update this line now that it is not a `FROM` statement?

tariq1890: I am not sure. It depends on dependabot's Dockerfile provider I suppose

tariq1890: On doing a cursory lookup, it seems unlikely that dependabot would update any docker image reference other than one in a `FROM` statement. We'd have to use renovate to automate updates

tariq1890: I'll add the renovate job in a follow-up.

PR #418, opened 2025-08-26
[ubuntu24.04][FIPS] disable FIPS enforcement temporarily when enabling pro mode

tariq1890:
When running `pro attach` on an Ubuntu 24.04 container, we are faced with a strange "chicken & egg" problem as the pro client attempts to use openssl + fips when it detects that the underlying kernel is fips_enabled, however, the FIPS packages are only available after the pro client enables APT repo access to the FIPS packages.

To work around this problem, we explicitly set the env var `OPENSSL_FORCE_FIPS_MODE=0` to unblock the pro client calls. After successful enablement of FIPS mode, we download the OpenSSL FIPS module and unset the `OPENSSL_FORCE_FIPS_MODE` env var

### DIFF
--- a/ubuntu24.04/nvidia-driver
+++ b/ubuntu24.04/nvidia-driver
@@ -28,10 +28,20 @@ _enable_fips_if_required() {
     if [[ -n "${UBUNTU_PRO_TOKEN}" && ${KERNEL_VERSION} =~ "fips" ]]; then
       echo "Ubuntu Pro token and FIPS kernel detected"
       apt-get -qq install --no-install-recommends ubuntu-advantage-tools > /dev/null
+
+      # This workaround is needed in Ubuntu 24.04 as OpenSSL attempts to leverage the FIPS provider
+      # when the underlying kernel is FIPS enabled.
+      export OPENSSL_FORCE_FIPS_MODE=0
+
       echo "Attaching Ubuntu Pro token..."
       pro attach --no-auto-enable "${UBUNTU_PRO_TOKEN}"
       echo "Enabling FIPS apt repo access..."
       pro enable --access-only --assume-yes fips-updates
+      echo "Installing the OpenSSL FIPS modules"
+      apt-get -qq install --no-install-recommends ubuntu-fips-userspace > /dev/null
+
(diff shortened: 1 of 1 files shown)


cdesiniotis (approved):

ubuntu24.04/nvidia-driver
```diff
@@ -28,10 +28,18 @@ _enable_fips_if_required() {
...
+      # when the underlying kernel is FIPS enabled.
+      export OPENSSL_FORCE_FIPS_MODE=0
+
       echo "Attaching Ubuntu Pro token..."
       pro attach --no-auto-enable "${UBUNTU_PRO_TOKEN}"
       echo "Enabling FIPS apt repo access..."
       pro enable --access-only --assume-yes fips-updates
+      echo "Installing the OpenSSL FIPS module"
+      apt-get -qq install --no-install-recommends openssl-fips-module-3 > /dev/null
+      unset OPENSSL_FORCE_FIPS_MODE
```
tariq1890, line 44: @cdesiniotis I am wondering if I should do this instead
```suggestion
      export OPENSSL_FORCE_FIPS_MODE=1
```

tariq1890: I've also built a container image with the above change and tested it. Works well

tariq1890: Ok this comment adds a bit more clarity: https://github.com/dotnet/dotnet-docker/issues/5849#issuecomment-2326054391

tariq1890: What I gather from that comment is - OpenSSL in Ubuntu 24.04 will always look to use FIPS if the underlying kernel has `/proc/sys/crypto/fips_enabled` set to 1. So the lack of the `OPENSSL_FORCE_FIPS_MODE` env var will not affect this behaviour. Setting `OPENSSL_FORCE_FIPS_MODE=0` is our way to override this in-built behaviour of OpenSSL Ubuntu 24.04. 

Given the above, I feel more comfortable with the `unset` statement now

cdesiniotis (approved):

PR #427, opened 2025-09-05
Precompiled: disabled the ubuntu22.04 6.8 kernel

shivakunv:
disabled the ubuntu22.04 6.8 kernel as it was unnecessarily consuming AWS resource for e2e testing. we will enable it once we decide to release


shivakunv:
need to enable ubuntu22.04 6.8 kernel driver . closing.

PR #452, opened 2025-09-30
Bump 570.195.03, 535.274.02, and 580.95.05 remove ubuntu20.04

JunAr7112:


cdesiniotis:
Let's also bump 580 to 580.95.05 -- this was just released today.
https://docs.nvidia.com/datacenter/tesla/index.html

cdesiniotis:
@JunAr7112 also update the commit message -- it still does not mention R580

.common-ci.yml
```diff
@@ -81,19 +81,13 @@ trigger-pipeline:
 .driver-versions:
   parallel:
     matrix:
-      - DRIVER_VERSION: [535.261.03, 570.172.08, 580.82.07]
-
-# Define the driver versions for jobs that can be run in parallel
-.driver-versions-ubuntu20.04:
-  parallel:
-    matrix:
-      - DRIVER_VERSION: [535.261.03, 570.172.08]
```
cdesiniotis, line 90: Not a blocker -- But Is there a reason this was removed in the first commit? I was expecting to see this in the second commit that removes support for ubuntu20.04.

.common-ci.yml
cdesiniotis, line 1: Question -- I see you have removed ubuntu20.04 jobs in `.common-ci.yml`. Are there no ubuntu20.04 jobs that need to be removed from `.nvidia-ci.yml` also?

.common-ci.yml
JunAr7112: This should be in the second commit now

.common-ci.yml
JunAr7112: I think so, since we won't be releasing with it. Removing this as well.

PR #454, opened 2025-10-02
Add fast-track to skip uninstall/install if NVIDIA driver modules present

karthikvetrivel:
This PR is a part of this endeavor:

> GPU Driver container should avoid re-installing drivers on spurious container restarts

Relevant PRs: 

- https://github.com/NVIDIA/gpu-operator/pull/1746
- https://github.com/NVIDIA/k8s-driver-manager/pull/103

### DIFF
--- a/rhel9/common.sh
+++ b/rhel9/common.sh
@@ -53,14 +53,15 @@ _read_conf_file() {
 }
 
 # Build driver configuration for state comparison
+# Note: Variables are expected to be set by the sourcing script (nvidia-driver)
 _build_driver_config() {
     cat <<EOF
 DRIVER_VERSION=${DRIVER_VERSION}
-DRIVER_TYPE=${DRIVER_TYPE:-passthrough}
+DRIVER_TYPE=${DRIVER_TYPE}
 KERNEL_VERSION=$(uname -r)
-GPU_DIRECT_RDMA_ENABLED=${GPU_DIRECT_RDMA_ENABLED:-false}
-USE_HOST_MOFED=${USE_HOST_MOFED:-false}
-KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE:-auto}
+GPU_DIRECT_RDMA_ENABLED=${GPU_DIRECT_RDMA_ENABLED}
+USE_HOST_MOFED=${USE_HOST_MOFED}
+KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE}
 NVIDIA_MODULE_PARAMS=$(_read_conf_file /drivers/nvidia.conf)
 NVIDIA_UVM_MODULE_PARAMS=$(_read_conf_file /drivers/nvidia-uvm.conf)
 NVIDIA_MODESET_MODULE_PARAMS=$(_read_conf_file /drivers/nvidia-modeset.conf)
@@ -70,8 +71,8 @@ EOF
 
 # Check if fast path should be used (driver already loaded with matching config)
 _should_use_fast_path() {
-    [ -f /sys/module/nvidia/refcnt ] && [ -f /run/nvidia/driver-config.state ] || return 1
+    [ -f /sys/module/nvidia/refcnt ] && [ -f /run/nvidia/nvidia-driver.state ] || return 1
     local current_config=$(_build_driver_config)
-    local stored_config=$(cat /run/nvidia/driver-config.state 2>/dev/null || echo "")
+    local stored_config=$(cat /run/nvidia/nvidia-driver.state 2>/dev/null || echo "")
     [ "${current_config}" = "${stored_config}" ]
 }
--- a/rhel9/nvidia-driver
+++ b/rhel9/nvidia-driver
@@ -8,6 +8,7 @@ PID_FILE=${RUN_DIR}/${0##*/}.pid
 DRIVER_VERSION=${DRIVER_VERSION:?"Missing DRIVER_VERSION env"}
 KERNEL_UPDATE_HOOK=/run/kernel/postinst.d/update-nvidia-driver
 NUM_VGPU_DEVICES=0
+DRIVER_TYPE="${DRIVER_TYPE:-passthrough}"
 GPU_DIRECT_RDMA_ENABLED="${GPU_DIRECT_RDMA_ENABLED:-false}"
 USE_HOST_MOFED="${USE_HOST_MOFED:-false}"
 NVIDIA_MODULE_PARAMS=()
@@ -212,10 +213,7 @@ _create_driver_package() (
     local nvidia_modeset_sign_args=""
     local nvidia_uvm_sign_args=""
 
-    # Skip cleanup trap for DTK builds - modules are copied after this function returns
-    if [ "${PACKAGE_TAG:-}" != "builtin" ]; then
-        trap "make -s -j ${MAX_THREADS} SYSSRC=/lib/modules/${KERNEL_VERSION}/build clean > /dev/null" EXIT
-    fi
+    trap "make -s -j ${MAX_THREADS} SYSSRC=/lib/modules/${KERNEL_VERSION}/build clean > /dev/null" EXIT
 
     echo "Compiling NVIDIA driver kernel modules..."
     cd /usr/src/nvidia-${DRIVER_VERSION}/${KERNEL_TYPE}
@@ -444,6 +442,21 @@ _unload_driver() {
         fi
     fi
 
+    if [ -f /var/run/nvidia-topologyd/nvidia-topologyd.pid ]; then
+        echo "Stopping NVIDIA topology daemon..."
+        local pid=$(< /var/run/nvidia-topologyd/nvidia-topologyd.pid)
+
+        kill -SIGTERM "${pid}"
+        for i in $(seq 1 50); do
+            kill -0 "${pid}" 2> /dev/null || break
+            sleep 0.1
+        done
+        if [ $i -eq 50 ]; then
+            echo "Could not stop NVIDIA topology daemon" >&2
+            return 1
+        fi
+    fi
+
     if [ -f /var/run/nvidia-fabricmanager/nv-fabricmanager.pid ]; then
         echo "Stopping NVIDIA fabric manager daemon..."
         local pid=$(< /var/run/nvidia-fabricmanager/nv-fabricmanager.pid)
@@ -664,19 +677,6 @@ _start_vgpu_topology_daemon() {
     nvidia-topologyd
 }
 
-_ensure_persistence() {
-    local pid_file=/var/run/nvidia-persistenced/nvidia-persistenced.pid pid
-    if pid=$(<"${pid_file}" 2>/dev/null) && [ -n "${pid}" ] && kill -0 "${pid}" 2>/dev/null; then
-        return 0
-    fi
-
-    if command -v nvidia-persistenced >/dev/null 2>&1; then
-        nvidia-persistenced --persistence-mode || true
-    else
-        echo "nvidia-persistenced not found; continuing without persistence"
-    fi
-}
-
 _start_daemons() {
     echo "Starting NVIDIA persistence daemon..."
     nvidia-persistenced --persistence-mode
@@ -719,7 +719,7 @@ _start_daemons() {
 }
 
 _store_driver_config() {
-    local config_file="/run/nvidia/driver-config.state"
+    local config_file="${RUN_DIR}/nvidia-driver.state"
     echo "Storing driver configuration state..."
     _build_driver_config > "$config_file"
     echo "Driver configuration stored at $config_file"
@@ -737,7 +737,6 @@ _wait_for_signal() {
 _userspace_only_install() {
     echo "Detected matching loaded driver & config (${DRIVER_VERSION}); performing userspace-only install"
     _unmount_rootfs
-    _update_package_cache
 
     cd /drivers
     [ ! -d "NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}" ] && sh NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}.run -x
--- a/ubuntu22.04/nvidia-driver
+++ b/ubuntu22.04/nvidia-driver
@@ -8,6 +8,7 @@ PID_FILE=${RUN_DIR}/${0##*/}.pid
 DRIVER_VERSION=${DRIVER_VERSION:?"Missing DRIVER_VERSION env"}
 KERNEL_UPDATE_HOOK=/run/kernel/postinst.d/update-nvidia-driver
 NUM_VGPU_DEVICES=0
+DRIVER_TYPE="${DRIVER_TYPE:-passthrough}"
 GPU_DIRECT_RDMA_ENABLED="${GPU_DIRECT_RDMA_ENABLED:-false}"
 USE_HOST_MOFED="${USE_HOST_MOFED:-false}"
 NVIDIA_MODULE_PARAMS=()
@@ -344,44 +345,7 @@ _load_driver() {
         set +o xtrace -o nounset
     fi
 
-    echo "Starting NVIDIA persistence daemon..."
-    nvidia-persistenced --persistence-mode
-
-    if [ "${DRIVER_TYPE}" = "vgpu" ]; then
-        echo "Copying gridd.conf..."
-        cp /drivers/gridd.conf /etc/nvidia/gridd.conf
-        if [ "${VGPU_LICENSE_SERVER_TYPE}" = "NLS" ]; then
-            echo "Copying ClientConfigToken..."
-            mkdir -p  /etc/nvidia/ClientConfigToken/
-            cp /drivers/ClientConfigToken/* /etc/nvidia/ClientConfigToken/
-        fi
-
-        echo "Starting nvidia-gridd.."
-        LD_LIBRARY_PATH=/usr/lib/$DRIVER_ARCH-linux-gnu/nvidia/gridd nvidia-gridd
-
-        # Start virtual topology daemon
-        _start_vgpu_topology_daemon
-    fi
-
-    if _assert_nvlink5_system; then
-        _ensure_nvlink5_prerequisites || return 1
-        echo "Starting NVIDIA fabric manager daemon for NVLink5+..."
-
-        fm_config_file=/usr/share/nvidia/nvswitch/fabricmanager.cfg
-        fm_pid_file=/var/run/nvidia-fabricmanager/nv-fabricmanager.pid
-        nvlsm_config_file=/usr/share/nvidia/nvlsm/nvlsm.conf
-        nvlsm_pid_file=/var/run/nvidia-fabricmanager/nvlsm.pid
-        /usr/bin/nvidia-fabricmanager-start.sh --mode start \
-                                               --fm-config-file $fm_config_file \
-                                               --fm-pid-file $fm_pid_file \
-                                               --nvlsm-config-file $nvlsm_config_file \
-                                               --nvlsm-pid-file $nvlsm_pid_file
-
-    # If not a NVLink5+ switch, check for the presence of NVLink4 (or below) switches
-    elif _assert_nvswitch_system; then
-        echo "Starting NVIDIA fabric manager daemon..."
-        nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
-    fi
+    _start_daemons
 
     return 0
 }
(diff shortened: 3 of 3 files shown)


ubuntu22.04/nvidia-driver
```diff
@@ -638,11 +639,128 @@ _start_vgpu_topology_daemon() {
...
+		echo "nvidia-persistenced not found; continuing without persistence"
+	fi
+}
+
+_build_driver_config() {
+	local config="DRIVER_VERSION=${DRIVER_VERSION}
+KERNEL_VERSION=$(uname -r)
+GPU_DIRECT_RDMA_ENABLED=${GPU_DIRECT_RDMA_ENABLED}
+USE_HOST_MOFED=${USE_HOST_MOFED}
+KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE}"
```
tariq1890: We would need to record these values in the driver config as well

```
NVIDIA_MODULE_PARAMS
NVIDIA_UVM_MODULE_PARAMS
NVIDIA_MODESET_MODULE_PARAMS
NVIDIA_PEERMEM_MODULE_PARAMS
```

tariq1890:
Thanks @karthikvetrivel, this looks very promising! I was curious as to why `ubuntu 24.04` wasn't included in the changeset

ubuntu22.04/nvidia-driver
```diff
@@ -638,11 +639,128 @@ _start_vgpu_topology_daemon() {
...
+		# Copy the kernel module sources for sidecar containers (gdrcopy, nvidia-fs, etc.)
+		mkdir -p /usr/src/nvidia-${DRIVER_VERSION} && \
+			cp -r LICENSE mkprecompiled ${KERNEL_TYPE} /usr/src/nvidia-${DRIVER_VERSION}/ && \
+			sed '9,${/^\(kernel\|LICENSE\)/!d}' .manifest > /usr/src/nvidia-${DRIVER_VERSION}/.manifest
+			
+		# Mount the driver rootfs to make components available
+		_mount_rootfs
+		
+		# Ensure persistence daemon is running
+		_ensure_persistence_running
```
tariq1890: The name of this method can be shortened to this
```suggestion
		_ensure_persistenced
```

ubuntu22.04/nvidia-driver
karthikvetrivel: Updated.

ubuntu22.04/nvidia-driver
karthikvetrivel: Updated.

rhel9/nvidia-driver
```diff
@@ -566,7 +570,9 @@ _install_driver() {
         install_args+=("--skip-module-load")
     fi
 
-    IGNORE_CC_MISMATCH=1 nvidia-installer --kernel-module-only --no-drm --ui=none --no-nouveau-check -m=${KERNEL_TYPE} ${install_args[@]+"${install_args[@]}"}
+    # Prevent prompts when modules are already loaded (common in DTK context).
+    # Pipe "1" to auto-answer "Continue installation" when prompted about loaded modules.
+    echo "1" | IGNORE_CC_MISMATCH=1 nvidia-installer --kernel-module-only --no-drm --ui=none --no-nouveau-check -m=${KERNEL_TYPE} ${install_args[@]+"${install_args[@]}"}
```
tariq1890: do we need to pipe 1? I believe you can add `-s/--silent` to disable all prompting

rhel9/nvidia-driver
```diff
@@ -701,6 +707,114 @@ _start_vgpu_topology_daemon() {
...
+		nvidia_uvm_params=$(cat "/drivers/nvidia-uvm.conf" | tr '\n' ' ')
+	fi
+	if [ -f "/drivers/nvidia-modeset.conf" ]; then
+		nvidia_modeset_params=$(cat "/drivers/nvidia-modeset.conf" | tr '\n' ' ')
+	fi
+	if [ -f "/drivers/nvidia-peermem.conf" ]; then
+		nvidia_peermem_params=$(cat "/drivers/nvidia-peermem.conf" | tr '\n' ' ')
+	fi
+	
+	local config="DRIVER_VERSION=${DRIVER_VERSION}
```
cdesiniotis: Should we also include `DRIVER_TYPE`? This is not defined at the top of this file, but this does get set as build time. By default it is set to `passthrough`. We set it to `vgpu` when building vGPU guest driver containers.

rhel9/nvidia-driver
karthikvetrivel: Updated to --silent! I was aware of the flag and can't remember why I pipe'd 1, but I believe it works as well.

rhel9/nvidia-driver
```diff
@@ -701,6 +701,114 @@ _start_vgpu_topology_daemon() {
...
+    # Copy kernel module sources if not already present (needed for other containers)
+    if [ ! -d "/usr/src/nvidia-${DRIVER_VERSION}" ]; then
+        _resolve_kernel_type || exit 1
+        mkdir -p /usr/src/nvidia-${DRIVER_VERSION}
+        cp -r LICENSE mkprecompiled ${KERNEL_TYPE} /usr/src/nvidia-${DRIVER_VERSION}/
+        sed '9,${/^\(kernel\|LICENSE\)/!d}' .manifest > /usr/src/nvidia-${DRIVER_VERSION}/.manifest
+    fi
+    
+    _mount_rootfs
+    _ensure_persistence
```
cdesiniotis: What about the other userspace services / daemons? For example, in `_load_driver()` we start nvidia-gridd for the vGPU use case and fabricmanager for NVSwitch-based systems. I would assume we need to (re)start those in this "fast-track" driver install path.

What if we factored out the code in `_load_driver()` that starts daemons into its own function, i.e. `_start_daemons()`? We can invoke that here and in the top-level `init()` for the non "fast-track" driver install path. Thoughts?

rhel9/ocp_dtk_entrypoint
```diff
@@ -10,6 +10,50 @@ echo "Running $*"
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 source $SCRIPT_DIR/common.sh
 
+_build_driver_config() {
```
cdesiniotis: Instead of re-implementing this here, what about moving this to `common.sh` and both the `nvidia-driver` and `ocp_dtk_entrypoint` scripts can re-use it? The same would apply to the `_should_use_fast_path` method.

rhel9/ocp_dtk_entrypoint
```diff
@@ -18,6 +62,13 @@ nv-ctr-run-with-dtk() {
         exec bash -x nvidia-driver init
     fi
 
+    if _should_use_fast_path; then
+        echo "Fast path detected: skipping DTK build and module copy, proceeding with userspace-only install"
```
cdesiniotis, line 22: I don't see us updating the `dtk-build-driver()` function at all. What does the DTK container do in this case? Can you share logs from it?

rhel9/nvidia-driver
karthikvetrivel: Added DRIVER_TYPE.

rhel9/nvidia-driver
karthikvetrivel: Agreed! Updated.

rhel9/ocp_dtk_entrypoint
karthikvetrivel: Updated.

rhel9/ocp_dtk_entrypoint
karthikvetrivel: Good catch--this change was still staged for some reason. I pushed this change up. We see these two key logs in `nvidia-driver-ctr`:

- `Fast path detected: skipping DTK build and module copy, proceeding with userspace-only install`
- `Detected matching loaded driver & config (580.82.07); performing userspace-only install`

It should be good now.

ubuntu22.04/nvidia-driver
```diff
@@ -638,11 +639,148 @@ _start_vgpu_topology_daemon() {
...
+			echo "Detected matching loaded driver & config (${DRIVER_VERSION}); performing userspace-only install"
+			
+			# Skip kernel module unload since they're already loaded with correct version
+			# Unmount any existing rootfs
+			_unmount_rootfs
+			
+			# Update package cache for userspace install
+			_update_package_cache
+			_resolve_kernel_version || exit 1
+			_install_prerequisites
```
shivamerla: Can we skip this for user space only install. That way we don't have to keep pulling the kernel headers and devel packages on restart.

rhel9/common.sh
```diff
@@ -45,3 +45,33 @@ _gdrcopy_enabled() {
...
+
+# Build driver configuration for state comparison
+_build_driver_config() {
+    cat <<EOF
+DRIVER_VERSION=${DRIVER_VERSION}
+DRIVER_TYPE=${DRIVER_TYPE:-passthrough}
+KERNEL_VERSION=$(uname -r)
+GPU_DIRECT_RDMA_ENABLED=${GPU_DIRECT_RDMA_ENABLED:-false}
+USE_HOST_MOFED=${USE_HOST_MOFED:-false}
+KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE:-auto}
```
cdesiniotis: Question -- should we define a default value for this variable (and others) here or just rely on the default defined in `nvidia-driver`?

rhel9/nvidia-driver
```diff
@@ -211,7 +212,10 @@ _create_driver_package() (
     local nvidia_modeset_sign_args=""
     local nvidia_uvm_sign_args=""
 
-    trap "make -s -j ${MAX_THREADS} SYSSRC=/lib/modules/${KERNEL_VERSION}/build clean > /dev/null" EXIT
+    # Skip cleanup trap for DTK builds - modules are copied after this function returns
+    if [ "${PACKAGE_TAG:-}" != "builtin" ]; then
```
cdesiniotis: Question - could you elaborate on why this change is needed? What exactly does the `clean` make target do?

rhel9/nvidia-driver
```diff
@@ -701,6 +664,105 @@ _start_vgpu_topology_daemon() {
...
+
+    # If not a NVLink5+ switch, check for the presence of NVLink4 (or below) switches
+    elif _assert_nvswitch_system; then
+        echo "Starting NVIDIA fabric manager daemon..."
+        nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
+    fi
+}
+
+_store_driver_config() {
+    local config_file="/run/nvidia/driver-config.state"
```
cdesiniotis: nit - we could leverage the `RUN_DIR` variable defined
```suggestion
    local config_file="${RUN_DIR}/driver-config.state"
```

rhel9/nvidia-driver
```diff
@@ -701,6 +664,105 @@ _start_vgpu_topology_daemon() {
     nvidia-topologyd
 }
 
+_ensure_persistence() {
```
cdesiniotis: Where is this being called? I don't see this method being invoked in the rhel9 scripts.

ubuntu22.04/nvidia-driver
```diff
@@ -638,36 +639,87 @@ _start_vgpu_topology_daemon() {
...
+
+_read_conf_file() {
+    local file="$1"
+    [ -f "$file" ] && tr '\n' ' ' < "$file"
+}
+
+_build_driver_config() {
+    cat <<EOF
+DRIVER_VERSION=${DRIVER_VERSION}
+DRIVER_TYPE=${DRIVER_TYPE:-passthrough}
```
cdesiniotis: Similar to a comment I made in the rhel9 scripts, it might be preferred to define all the variables (and their default values) at the top of this file.

ubuntu22.04/nvidia-driver
```diff
@@ -681,8 +733,41 @@ init() {
...
+    if [ -f /sys/module/nvidia/refcnt ] && [ -f /run/nvidia/driver-config.state ]; then
+        current_config=$(_build_driver_config)
+        stored_config=$(cat /run/nvidia/driver-config.state)
+
+        if [ "${current_config}" = "${stored_config}" ]; then
+            echo "Detected matching loaded driver & config (${DRIVER_VERSION}); performing userspace-only install"
+            _unmount_rootfs
+            _update_package_cache
+            _resolve_kernel_version || exit 1
+            _install_prerequisites
```
cdesiniotis: Question -- is installing prerequisite packages actually required in this case? If yes, which packages are required for the userspace-only install?

ubuntu22.04/nvidia-driver
```diff
@@ -681,8 +733,41 @@ init() {
...
+        fi
+    fi
+
+    # Full install path: unload existing driver and perform complete installation
     _unload_driver || exit 1
     _unmount_rootfs
+    _install_userspace_components
+    _resolve_kernel_type || exit 1
+
+    # Move (not copy) kernel module sources since this is the full install path
```
cdesiniotis: Question -- could you explain why we copy in the userspace-only install path, but move in this code path?

rhel9/nvidia-driver
karthikvetrivel: To my understanding, make clean removes the compiled .ko module files. For DTK builds, the modules need to persist after `_create_driver_package()` returns because `ocp_dtk_entrypoint` copies them to a shared volume afterward. If the cleanup trap runs, the modules are deleted before the copy happens. 

Feel free to push back on my understanding here.

ubuntu22.04/nvidia-driver
karthikvetrivel: Originally, I thought that userspace-only path should use `cp` because the source directory may still be needed—for example, if the container restarts and re-enters the fast path, having the original extraction intact avoids potential issues. 

However, on second look, we extract (`sh NVIDIA-Linux-${DRIVER_ARCH}-${DRIVER_VERSION}.run -x`) every time so this can be updated to a `mv`. Thanks for pointing this out.

ubuntu22.04/nvidia-driver
karthikvetrivel: You're right, the prerequisites (kernel headers, kernel image, kernel modules) are specifically for kernel module compilation, which doesn't happen here—we run `nvidia-installer --no-kernel-module`. Will remove.

rhel9/nvidia-driver
karthikvetrivel: Updated.

rhel9/nvidia-driver
karthikvetrivel: Updated, replaced a prior `_start_daemons` method.

ubuntu22.04/nvidia-driver
karthikvetrivel: Updated. I have the variable defaults defined at the top of the file--let me know if thats what you intended.

rhel9/common.sh
karthikvetrivel: Updated to use the default defined in `nvidia-driver`

rhel9/nvidia-driver
```diff
@@ -701,6 +665,105 @@ _start_vgpu_topology_daemon() {
...
+    # Copy kernel module sources if not already present (needed for sidecar containers)
+    if [ ! -d "/usr/src/nvidia-${DRIVER_VERSION}" ]; then
+        _resolve_kernel_type || exit 1
+        mkdir -p /usr/src/nvidia-${DRIVER_VERSION}
+        cp -r LICENSE mkprecompiled ${KERNEL_TYPE} /usr/src/nvidia-${DRIVER_VERSION}/
+        sed '9,${/^\(kernel\|LICENSE\)/!d}' .manifest > /usr/src/nvidia-${DRIVER_VERSION}/.manifest
+    fi
+
+    _mount_rootfs
+    _ensure_persistence
```
cdesiniotis: Why are we only starting the `nvidia-persistenced` daemon for this code path? Shouldn't we be restarting all of the daemons that we bring up in `_start_daemons()`?

My intuition tells me that none of the daemons should still be running at this point -- that is, none of the daemons should still be running after a previously running driver container terminates / restarts. If my intuition is correct, then couldn't we unconditionally call `_start_daemons()` here? Let me know if you have observed different behavior.

ubuntu22.04/nvidia-driver
cdesiniotis: Let's also remove the call to `_update_package_cache` while we are at it. If we aren't installing any new packages then we don't need to update the package cache.

karthikvetrivel: Updated. Also removed in RHEL9.

rhel9/nvidia-driver
karthikvetrivel: i think you're right. Whenever we hit `_ensure_persistence` in this code path, we log persistence not found. I'll start the daemons here (and mirror the changes to Ubuntu 22.04).

Speaking of daemons, but I noticed that we don't ever stop `nvidia-topologyd` in `_unload_driver()` but we stop every other daemon. Is there a reason why?

cdesiniotis: > Speaking of daemons, but I noticed that we don't ever stop nvidia-topologyd in _unload_driver() but we stop every other daemon. Is there a reason why?

No reason AFAIK. We should also stop it in `_unload_driver()`.

rhel9/nvidia-driver
cdesiniotis: Are you suggesting that this change is needed in general and is not related to the fast-track install code path?

karthikvetrivel: Yes, this isn't related to a fast track optimization. This path isn't hit on a fast track reinstall. I can move these changes to a follow-up PR if preferred!

cdesiniotis: Yes, I would prefer if this change (and any other changes not related to the fast-track optimization) were in a separate PR with a clear description of the problem addressed.

karthikvetrivel: Removed.

rhel9/nvidia-driver
cdesiniotis, line 1: The permissions of this file have been changed.

rhel9/nvidia-driver
```diff
@@ -701,6 +677,91 @@ _start_vgpu_topology_daemon() {
...
+    [ "${ACCEPT_LICENSE}" = "yes" ] && install_args="$install_args --accept-license"
+    IGNORE_CC_MISMATCH=1 ./nvidia-installer $install_args
+
+    # Copy kernel module sources if not already present (needed for sidecar containers)
+    if [ ! -d "/usr/src/nvidia-${DRIVER_VERSION}" ]; then
+        _resolve_kernel_type || exit 1
+        mkdir -p /usr/src/nvidia-${DRIVER_VERSION}
+        cp -r LICENSE mkprecompiled ${KERNEL_TYPE} /usr/src/nvidia-${DRIVER_VERSION}/
+        sed '9,${/^\(kernel\|LICENSE\)/!d}' .manifest > /usr/src/nvidia-${DRIVER_VERSION}/.manifest
+    fi
```
cdesiniotis, line 756: Question -- isn't this already performed in the `_prepare` function? Could we just replace this with a call to `_prepare` or is there a nuance I am missing?

rhel9/nvidia-driver
```diff
@@ -758,17 +819,48 @@ _load() {
...
+
+    exec 3> ${PID_FILE}
+    if ! flock -n 3; then
+        echo "An instance of the NVIDIA driver is already running, aborting"
+        exit 1
+    fi
+    echo $$ >&3
+
+    trap "echo 'Caught signal'; exit 1" HUP INT QUIT PIPE TERM
+    trap "_shutdown" EXIT
```
cdesiniotis, line 842: To make this a bit easier to read, what if we reverted this change and continued to just call `_prepare_exclusive`? We can remove the following lines from `_prepare_exclusive` and put them at the call site (like you have already done below).
```
_unload_driver || exit 1
_unmount_rootfs
```
Feel free to push back if this complicates things.

rhel9/nvidia-driver
cdesiniotis: I am thinking this method would just turn into:
```
_userspace_only_install() {
    echo "Detected matching loaded driver & config (${DRIVER_VERSION}); performing userspace-only install"
    _unmount_rootfs
    _mount_rootfs
    _start_daemons
    _write_kernel_update_hook
    _store_driver_config
    echo "Userspace-only install complete"
}
```

cdesiniotis:
Closing in favor of https://github.com/NVIDIA/gpu-driver-container/pull/534

PR #457, opened 2025-10-04
[CI] use nv-hosted action runners for image builds

tariq1890:


.github/workflows/image.yaml
```diff
@@ -16,21 +16,15 @@
...
-    types:
-      - opened
-      - synchronize
-    branches:
-      - main
-      - release-*
   push:
     branches:
       - main
+      - "pull-request/[0-9]+"
```
cdesiniotis, line 22: Can you rebase? I believe this should already be updated on main.

cdesiniotis (approved):
one small comment, but lgtm otherwise.

ArangoGutierrez (approved):
++ on @cdesiniotis comment

.github/workflows/image.yaml
tariq1890: Rebased. I've also chosen to maintain lexical order here

ArangoGutierrez (approved):
Thanks for the rebase @tariq1890

PR #459, opened 2025-10-06
Update workflow to publish to GHCR on PR tests

JunAr7112:


.github/workflows/image.yaml
```diff
@@ -25,6 +25,7 @@ on:
       - release-*
   push:
     branches:
+      - "pull-request/[0-9]+"
```
cdesiniotis, line 21: Can we align the entire `on` block with what we have in other repos? For example https://github.com/NVIDIA/gpu-operator/blob/ee9210bbb64a1c6a01ac67f2d592210c6934a09f/.github/workflows/ci.yaml#L17-L22.

cdesiniotis: I don't have a strong opinion here, but I lean towards using a config that already works in other projects. I want to make sure the workflow status from the `pull-request/[0-9]*` branches is reported back to the original PRs.

JunAr7112: Ok. I can try that out.

cdesiniotis:
> /ok to test https://github.com/NVIDIA/gpu-driver-container/commit/44696536b98169a09563c5a7a6ca02391869cca4

This appeared to trigger the workflow and the status of the worklfow is viewable from this PR.

PR #469, opened 2025-10-15
add vgpu manager scripts for ubuntu24.04

JunAr7112:


vgpu-manager/ubuntu24.04/nvidia-driver
```diff
@@ -0,0 +1,338 @@
...
+  _load_driver || exit 1
+  _mount_rootfs
+  _enable_vfs
+
+  # In certain scenarios, /sys/class/mdev_bus is not populated with the correct list of devices (PFs and possible VFs)  at this point.
+  # Re-run nvdidia-vgpud to ensure /sys/class/mdev_bus is populated correctly. And restart nvidia-vgpu-mgr if previously killed.
+  nvidia-vgpud &
+  pgrep nvidia-vgpu-mgr >/dev/null || (echo "Restarting nvidia-vgpu-mgr after previously killed" && nvidia-vgpu-mgr &)
+
+  # Create validation file to indicate vGPU manager is ready
```
JunAr7112: I didn't see a step to create a validation file in the ubuntu20.04 and other repos; however, I was running into an error without creating a validation file in nvidia-driver entrypoint.

vgpu-manager/ubuntu24.04/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM nvcr.io/nvidia/cuda:13.0.1-base-ubuntu24.04
```
tariq1890:
```suggestion
FROM nvcr.io/nvidia/cuda:13.0.2-base-ubuntu24.04
```

vgpu-manager/ubuntu24.04/nvidia-driver
tariq1890: We shouldn't be adding this here. It's a container in the gpu-operator-validator pod that is responsible for creating this sentinel file

JunAr7112: Ok I can remove this from here.

vgpu-manager/ubuntu24.04/Dockerfile
JunAr7112: Alright I updated the version @tariq1890

vgpu-manager/ubuntu24.04/nvidia-driver
```diff
@@ -0,0 +1,333 @@
...
+        fi
+    fi
+    return 0
+}
+
+_shutdown() {
+    if _disable_vfs && _unload_driver; then
+        _unmount_rootfs
+        # Remove validation file
+        rm -f /run/nvidia/validations/vgpu-manager-ready
```
tariq1890: shouldn't this be removed too?

tariq1890: I don't see this in the `nvidia-driver` file under `ubuntu22.04`

JunAr7112: I think its safe to remove this. I added it earlier while testing, but I just checked and it shouldn't be needed for deploying.

PR #512, opened 2025-12-15
gpu-manager: enable kernel module configuration via KernelModuleConfig

shivakunv:


vgpu-manager/rhel8/nvidia-driver
```diff
@@ -52,14 +54,59 @@ _set_fw_search_path() {
...
+        echo "NVIDIA modules already loaded by installer, unloading to apply custom parameters..."
+        rmmod nvidia_vgpu_vfio 2>/dev/null || true
+        rmmod nvidia 2>/dev/null || true
+    fi
+
+    echo "Loading NVIDIA driver kernel modules..."
+    set -o xtrace +o nounset
+    modprobe nvidia
+    modprobe nvidia_vgpu_vfio
+    set +o xtrace -o nounset
```
cdesiniotis: This change did not make sense to me at first. But then I realized that the `_install_driver()` actually loads the modules (we leverage the nvidia-installer to load the modules, in addition to building+installing the driver). 

Question -- If we create the modprobe conf file (`/etc/modprobe.d/nvidia.conf`) BEFORE calling `_install_driver()`, what is the observed behavior? Do the modules get loaded with the desired parameters?

shivakunv: 1.  _I  put the `else` condition inside `_create_module_params_conf` and tested it , it is working fine ._
```
 else
        echo "create nvidia.conf file"
        echo "options nvidia NVreg_EnableGpuFirmwareLogs=3" > ${MODPROBE_CONFIG_DIR}/nvidia.conf
    fi
```
```
_create_module_params_conf() {
    echo "Parsing kernel module parameters..."
    _get_module_params

    if [ ${#NVIDIA_MODULE_PARAMS[@]} -gt 0 ]; then
        echo "Configuring nvidia module parameters in ${MODPROBE_CONFIG_DIR}/nvidia.conf"
        echo "options nvidia ${NVIDIA_MODULE_PARAMS[@]}" > ${MODPROBE_CONFIG_DIR}/nvidia.conf
    else
        echo "create nvidia.conf file"
        echo "options nvidia NVreg_EnableGpuFirmwareLogs=3" > ${MODPROBE_CONFIG_DIR}/nvidia.conf
    fi
}

```
2. _I created a configmap for the nvdia.conf file and tested it , which is also working fine .  ( refer : PR: [https://github.com/NVIDIA/gpu-operator/pull/1946](https://github.com/NVIDIA/gpu-operator/pull/1946)_



If I need to test it in other ways, please list the steps .

cdesiniotis: My original comment is questioning whether this code snippet (lines 96-L107) is actually needed. Can you remove this snippet and verify whether custom module parameters can be applied or not?

shivakunv: It is needed . 
After removal, the custom module parameter does not take effect and required a system reboot  to be applied.

cdesiniotis: Okay, thanks for confirming. I still would like to explore this further since for the other driver containers (not vgpu-manager) this code snippet is not needed and custom module parameters are able to be applied correctly.

In `ubuntu24.04/nvidia-driver`, we set the `--skip-module-load` flag when installing the driver in `_instal_driver()` (see [here](https://github.com/NVIDIA/gpu-driver-container/blob/e4f05d3dae252bf1068452b61c86f9d779073f4e/ubuntu24.04/nvidia-driver#L470-L484)). Can you try the following:

1) Add `--skip-module-load` flag to `_install_driver()`
2) Remove this code snippet in `_load_driver()`
3) Check if you can apply custom module parameters.

shivakunv: Thanks!! working.
updated the code

PR #513, opened 2025-12-16
update lspci path in vgpu-manager for RHEL

akri3i:
This PR corrects the file path used to locate the `lspci` utility within the `vgpu-manager` entrypoint script (`ocp_dtk_entrypoint`).
Currently it tries to copy lscpi from /usr/bin
While on RHEL it is located in /usr/sbin

Tested this while building vgpu manager image 
The following error occurs during the nvidia-vgpu-manager-daemonset initialization
`+ cp -r /usr/local/bin/ocp_dtk_entrypoint /usr/local/bin/nvidia-driver /usr/bin/lspci /driver /mnt/shared-nvidia-driver-toolkit/
cp: cannot stat '/usr/bin/lspci': No such file or directory`

which is fixed when path is changed to /usr/sbin/lspci


akri3i:
@cdesiniotis

PR #527, opened 2025-12-17
Add Renovate to automate Version Bumps

JunAr7112:


.github/renovate.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": ["config:base"],
+    "enabledManagers": ["dockerfile"],
+    "schedule": ["daily"],
+    "labels": ["dependencies", "docker"],
+    "prConcurrentLimit": 15
+}
```
tariq1890: can you add an EOL here?

JunAr7112: Added line

.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,21 @@
+name: Renovate
+
+on:
+  schedule:
```
cdesiniotis, line 5: Let's add `workflow_dispatch` so that we can also manually trigger this action (in addition to the schedule)
```suggestion
on:
  workflow_dispatch:
  schedule:
```

JunAr7112: Alright

cdesiniotis (approved):
LGTM. let's get @tariq1890's review as well.

PR #528, opened 2025-12-22
add support for driver version 590.48.01

tariq1890:
This PR also includes a refactor in the `install.sh` files where we stop referring to the packages by their rpm/deb package versions explicitly

This simplifies the if conditionals signficantly and also ensures the latest rpm/deb package versions are always downloaded by relying on `dnf/apt`'s wildcard matching.


rhel10/install.sh
```diff
@@ -119,18 +119,42 @@ nvidia_installer () {
 }
 
 fabricmanager_install() {
+  if [ "$DRIVER_VERSION" = "590.48.01" ]; then
+    # This is temporary. fabricmanager 590.48.01 has an rpm package fix that we depend on, so we pull the "-2" package version
+    dnf install -y nvidia-fabricmanager-${DRIVER_VERSION}-2.el10
```
cdesiniotis: Question -- is there any way to just "install the latest" instead of having to hardcode the `-2` patch version? What happens if we do:
```
dnf install -y nvidia-fabricmanager-${DRIVER_VERSION}
```

cdesiniotis: Remind me again, why do we need the `.el10` suffix?

tariq1890: The packaging team has changed the rpm and deb package naming format starting with R590

tariq1890: > Question -- is there any way to just "install the latest" instead of having to hardcode the -2 patch version? 

Let me check

tariq1890: @cdesiniotis This has been addressed. Please check now

cdesiniotis (approved):

PR #496, opened 2025-12-02
Add support for rhel10.0 and rhel10.1

shivakunv:

PR #534, opened 2026-01-08
[GPU Driver Container Resiliency] Implement userspace-only fast path when kernel modules already loaded with matching config

karthikvetrivel:
Relevant PRs: 

- https://github.com/NVIDIA/gpu-operator/pull/2024
- https://github.com/NVIDIA/k8s-driver-manager/pull/148

## Description 

- Implements userspace-only fast path install when kernel modules are already loaded with matching configuration
- Checks if NVIDIA modules are loaded (`/sys/module/nvidia/refcnt`) and compares configuration digest
- When match detected, performs userspace-only install (libraries/binaries) while skipping kernel module build/load
- Stores digest after successful installation for future comparisons

### DIFF
--- a/rhel9/common.sh
+++ b/rhel9/common.sh
@@ -45,3 +45,13 @@ _gdrcopy_enabled() {
     fi
     return 1
 }
+
+# Check if fast path should be used (driver already loaded with matching config)
+# Compares current digest from DRIVER_CONFIG_DIGEST env var with stored digest
+_should_skip_kernel_module_reload() {
+    [ -f /sys/module/nvidia/refcnt ] && [ -f /run/nvidia/nvidia-driver.state ] || return 1
+    local current_digest="${DRIVER_CONFIG_DIGEST:-}"
+    [ -z "${current_digest}" ] && return 1
+    local stored_digest=$(cat /run/nvidia/nvidia-driver.state 2>/dev/null || echo "")
+    [ "${current_digest}" = "${stored_digest}" ]
+}
--- a/rhel9/nvidia-driver
+++ b/rhel9/nvidia-driver
@@ -418,44 +418,7 @@ _load_driver() {
         set +o xtrace -o nounset
     fi
 
-    echo "Starting NVIDIA persistence daemon..."
-    nvidia-persistenced --persistence-mode
-
-    if [ "${DRIVER_TYPE}" = "vgpu" ]; then
-        echo "Copying gridd.conf..."
-        cp /drivers/gridd.conf /etc/nvidia/gridd.conf
-        if [ "${VGPU_LICENSE_SERVER_TYPE}" = "NLS" ]; then
-            echo "Copying ClientConfigToken..."
-            mkdir -p  /etc/nvidia/ClientConfigToken/
-            cp /drivers/ClientConfigToken/* /etc/nvidia/ClientConfigToken/
-        fi
-
-        echo "Starting nvidia-gridd.."
-        LD_LIBRARY_PATH=/usr/lib64/nvidia/gridd nvidia-gridd
-
-        # Start virtual topology daemon
-        _start_vgpu_topology_daemon
-    fi
-
-    if _assert_nvlink5_system; then
-        _ensure_nvlink5_prerequisites || return 1
-        echo "Starting NVIDIA fabric manager daemon for NVLink5+..."
-
-        fm_config_file=/usr/share/nvidia/nvswitch/fabricmanager.cfg
-        fm_pid_file=/var/run/nvidia-fabricmanager/nv-fabricmanager.pid
-        nvlsm_config_file=/usr/share/nvidia/nvlsm/nvlsm.conf
-        nvlsm_pid_file=/var/run/nvidia-fabricmanager/nvlsm.pid
-        /usr/bin/nvidia-fabricmanager-start.sh --mode start \
-                                               --fm-config-file $fm_config_file \
-                                               --fm-pid-file $fm_pid_file \
-                                               --nvlsm-config-file $nvlsm_config_file \
-                                               --nvlsm-pid-file $nvlsm_pid_file
-
-    # If not a NVLink5+ switch, check for the presence of NVLink4 (or below) switches
-    elif _assert_nvswitch_system; then
-        echo "Starting NVIDIA fabric manager daemon..."
-        nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
-    fi
+    _start_daemons
 }
 
 # Stop persistenced and unload the kernel modules if they are currently loaded.
@@ -493,6 +456,21 @@ _unload_driver() {
         done
         if [ $i -eq 50 ]; then
             echo "Could not stop NVIDIA Grid daemon" >&2
+            return 1
+        fi
+    fi
+
+    if [ -f /var/run/nvidia-topologyd/nvidia-topologyd.pid ]; then
+        echo "Stopping NVIDIA topology daemon..."
+        local pid=$(< /var/run/nvidia-topologyd/nvidia-topologyd.pid)
+
+        kill -SIGTERM "${pid}"
+        for i in $(seq 1 50); do
+            kill -0 "${pid}" 2> /dev/null || break
+            sleep 0.1
+        done
+        if [ $i -eq 50 ]; then
+            echo "Could not stop NVIDIA topology daemon" >&2
             return 1
         fi
     fi
@@ -587,10 +565,6 @@ _install_driver() {
     fi
 
     IGNORE_CC_MISMATCH=1 nvidia-installer --kernel-module-only --no-drm --ui=none --no-nouveau-check -m=${KERNEL_TYPE} ${install_args[@]+"${install_args[@]}"}
-    # May need to add no-cc-check for Rhel, otherwise it complains about cc missing in path
-    # /proc/version and lib/modules/KERNEL_VERSION/proc are different, by default installer looks at /proc/ so, added the proc-mount-point
-    # TODO: remove the -a flag. its not needed. in the new driver version, license-acceptance is implicit
-    #nvidia-installer --kernel-module-only --no-drm --ui=none --no-nouveau-check --no-cc-version-check --proc-mount-point /lib/modules/${KERNEL_VERSION}/proc ${install_args[@]+"${install_args[@]}"}
 }
 
 # Mount the driver rootfs into the run directory with the exception of sysfs.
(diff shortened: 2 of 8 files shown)


rhel9/nvidia-driver
```diff
@@ -8,12 +8,13 @@ PID_FILE=${RUN_DIR}/${0##*/}.pid
 DRIVER_VERSION=${DRIVER_VERSION:?"Missing DRIVER_VERSION env"}
 KERNEL_UPDATE_HOOK=/run/kernel/postinst.d/update-nvidia-driver
 NUM_VGPU_DEVICES=0
+DRIVER_TYPE="${DRIVER_TYPE:-passthrough}"
+USE_HOST_MOFED="${USE_HOST_MOFED:-false}"
```
cdesiniotis: Can this change be reverted now that we moved to using `DRIVER_CONFIG_DIGEST`?

rhel9/nvidia-driver
```diff
@@ -586,11 +565,7 @@ _install_driver() {
         install_args+=("--skip-module-load")
     fi
 
-    IGNORE_CC_MISMATCH=1 nvidia-installer --kernel-module-only --no-drm --ui=none --no-nouveau-check -m=${KERNEL_TYPE} ${install_args[@]+"${install_args[@]}"}
-    # May need to add no-cc-check for Rhel, otherwise it complains about cc missing in path
-    # /proc/version and lib/modules/KERNEL_VERSION/proc are different, by default installer looks at /proc/ so, added the proc-mount-point
-    # TODO: remove the -a flag. its not needed. in the new driver version, license-acceptance is implicit
-    #nvidia-installer --kernel-module-only --no-drm --ui=none --no-nouveau-check --no-cc-version-check --proc-mount-point /lib/modules/${KERNEL_VERSION}/proc ${install_args[@]+"${install_args[@]}"}
+    IGNORE_CC_MISMATCH=1 nvidia-installer --silent --kernel-module-only --no-drm --ui=none --no-nouveau-check -m=${KERNEL_TYPE} ${install_args[@]+"${install_args[@]}"}
```
cdesiniotis: Not sure if we already raised a thread for this, but is the only change here the addition of the `--silent` flag? If yes, let's remove this change to minimize the diff here. If you want to make this change in a follow-up, feel free to do that.

rhel9/nvidia-driver
```diff
@@ -721,6 +696,73 @@ _start_vgpu_topology_daemon() {
...
+    echo "Done, now waiting for signal"
+    sleep infinity &
+    trap "echo 'Caught signal'; _shutdown && { kill $!; exit 0; }" HUP INT QUIT PIPE TERM
+    trap - EXIT
+    while true; do wait $! || continue; done
+    exit 0
+}
+
+_userspace_only_install() {
+    echo "Detected matching loaded driver & config (${DRIVER_VERSION}); performing userspace-only install"
```
cdesiniotis: nit:
```suggestion
    echo "The NVIDIA driver is already loaded with the desired configuration, performing userspace-only install"
```

rhel9/ocp_dtk_entrypoint
```diff
@@ -82,6 +89,19 @@ dtk-build-driver() {
         sleep inf
     fi
 
+    # Check if fast path is being used - if so, skip building and signal completion
+    if _should_use_fast_path; then
+        echo "Fast path detected in DTK container: driver already loaded with matching config, skipping build"
```
cdesiniotis:
```suggestion
        echo "The NVIDIA driver is already loaded with the desired configuration, skipping build"
```

rhel9/ocp_dtk_entrypoint
```diff
@@ -82,6 +89,19 @@ dtk-build-driver() {
...
+        echo "Signaling driver_built and sleeping forever..."
```
cdesiniotis:
```suggestion
        echo "Signaling driver_built to the main container and sleeping forever..."
```

rhel9/ocp_dtk_entrypoint
```diff
@@ -18,6 +18,13 @@ nv-ctr-run-with-dtk() {
         exec bash -x nvidia-driver init
     fi
 
+    if _should_use_fast_path; then
+        echo "Fast path detected: skipping DTK build and module copy, proceeding with userspace-only install"
```
cdesiniotis:
```suggestion
        echo "The NVIDIA driver is already loaded with the desired configuration, skipping kernel module build and proceeding with userspace-only install"
```

ubuntu22.04/nvidia-driver
```diff
@@ -8,6 +8,7 @@ PID_FILE=${RUN_DIR}/${0##*/}.pid
 DRIVER_VERSION=${DRIVER_VERSION:?"Missing DRIVER_VERSION env"}
 KERNEL_UPDATE_HOOK=/run/kernel/postinst.d/update-nvidia-driver
 NUM_VGPU_DEVICES=0
+DRIVER_TYPE="${DRIVER_TYPE:-passthrough}"
```
cdesiniotis: Can we remove this change now that we are using `DRIVER_CONFIG_DIGEST`?

ubuntu22.04/nvidia-driver
```diff
@@ -550,6 +529,7 @@ _mount_rootfs() {
     mount --make-private /sys
     mkdir -p ${RUN_DIR}/driver
     mount --rbind / ${RUN_DIR}/driver
+    echo "Driver container rootfs mounted at ${RUN_DIR}/driver"
```
cdesiniotis: Let's remove this change to minimize the diff.

ubuntu22.04/nvidia-driver
```diff
@@ -701,8 +748,34 @@ init() {
...
     trap "_shutdown" EXIT
 
+    # Fast path: if NVIDIA kernel modules are already loaded and digest matches,
+    # skip kernel module build/load and only reinstall userspace components.
+    # This handles non-clean restarts where modules are in use and can't be unloaded.
+    if [ -f /sys/module/nvidia/refcnt ] && [ -f /run/nvidia/nvidia-driver.state ]; then
+        current_digest=$(_get_current_digest)
+        stored_digest=$(_read_stored_digest)
+
+        if [ -n "${current_digest}" ] && [ "${current_digest}" = "${stored_digest}" ]; then
```
cdesiniotis: Similar to the changes you made in `rhel9`, can we create a helper function to check if the fast-path install path should be taken or not? That will simplify the call-site here.

rhel9/nvidia-driver
karthikvetrivel: Removed.

rhel9/nvidia-driver
karthikvetrivel: Yes. I also realized it is redundant with the `--ui=none` flag.

rhel9/nvidia-driver
karthikvetrivel: Updated.

rhel9/ocp_dtk_entrypoint
karthikvetrivel: Updated.

rhel9/ocp_dtk_entrypoint
karthikvetrivel: Updated.

rhel9/ocp_dtk_entrypoint
karthikvetrivel: Updated.

ubuntu22.04/nvidia-driver
karthikvetrivel: Removed.

ubuntu22.04/nvidia-driver
karthikvetrivel: Removed.

ubuntu22.04/nvidia-driver
karthikvetrivel: Updated the call-site, should look a bit cleaner here now.

rhel9/nvidia-driver
```diff
@@ -721,6 +695,73 @@ _start_vgpu_topology_daemon() {
...
+_wait_for_signal() {
+    echo "Done, now waiting for signal"
+    sleep infinity &
+    trap "echo 'Caught signal'; _shutdown && { kill $!; exit 0; }" HUP INT QUIT PIPE TERM
+    trap - EXIT
+    while true; do wait $! || continue; done
+    exit 0
+}
+
+_userspace_only_install() {
```
tariq1890: I believe this name would imply that it is "userspace only"
```suggestion
_userspace_install() {
```

rhel9/common.sh
```diff
@@ -45,3 +45,13 @@ _gdrcopy_enabled() {
     fi
     return 1
 }
+
+# Check if fast path should be used (driver already loaded with matching config)
+# Compares current digest from DRIVER_CONFIG_DIGEST env var with stored digest
+_should_use_fast_path() {
```
tariq1890: How about this as the method name?
```suggestion
_should_skip_kernel_module_reload() {
```

**OR** invert the conditional statement and use the name below

```suggestion
_should_rebuild_kernel_module() {
```

rhel9/nvidia-driver
karthikvetrivel: Updated.

rhel9/common.sh
karthikvetrivel: I don't have a strong preference either way, so updated.

rhel9/nvidia-driver
```diff
@@ -758,8 +797,7 @@ _prepare_exclusive() {
     trap "echo 'Caught signal'; exit 1" HUP INT QUIT PIPE TERM
     trap "_shutdown" EXIT
 
-    _unload_driver || exit 1
-    _unmount_rootfs
+    _prepare
```
cdesiniotis: Question -- was there a technical reason for moving the call to `_prepare` to the bottom of this function?

karthikvetrivel: No, I believe it should be all the same. I'll move it back up to minimize the diff.

rhel9/ocp_dtk_entrypoint
```diff
@@ -82,6 +89,19 @@ dtk-build-driver() {
...
     fi
 
+    # Check if fast path is being used - if so, skip building and signal completion
+    if _should_skip_kernel_module_reload; then
+        echo "The NVIDIA driver is already loaded with the desired configuration, skipping build"
+        echo "Signaling driver_built to the main container and sleeping forever..."
+        touch "$DRIVER_TOOLKIT_SHARED_DIR/driver_build_started"
+        touch "$DRIVER_TOOLKIT_SHARED_DIR/driver_built"
+        while [ -f "$DRIVER_TOOLKIT_SHARED_DIR/driver_built" ]; do
+            sleep 30
```
tariq1890: What's the reason for the 30 second sleep?

karthikvetrivel: The `sleep 30` is just a keep-alive loop that checks every 30 seconds if the container should still be running.

ubuntu24.04/nvidia-driver
```diff
@@ -604,6 +678,19 @@ init() {
     trap "echo 'Caught signal'; exit 1" HUP INT QUIT PIPE TERM
     trap "_shutdown" EXIT
 
+    if _should_skip_kernel_module_reload; then
+        echo "The NVIDIA driver is already loaded with the desired configuration, performing userspace-only install"
+        _unmount_rootfs
+        _install_userspace_components
+        _mount_rootfs
```
cdesiniotis: The one difference I see between `ubuntu22.04` and the implementation here for `ubuntu24.04` is that `ubuntu22.04` has the below while `ubuntu24.04` does not:
```
_resolve_kernel_type || exit 1
_move_kernel_module_sources
```
While `_resolve_kernel_type` is not a current helper function in this file, we still need to move the kernel module sources right? I would assume making the source available is required if additional modules (e.g. gdrdrv, nvidia_fs) need to be compiled.

cdesiniotis: I don't see why we couldn't introduce the `_resolve_kernel_type` helper in this file as is implemented here:  https://github.com/NVIDIA/gpu-driver-container/blob/e4f05d3dae252bf1068452b61c86f9d779073f4e/ubuntu22.04/nvidia-driver#L593C1-L617C2

karthikvetrivel: Introduced to Ubuntu 24.04.

ubuntu24.04/nvidia-driver
```diff
@@ -604,6 +707,21 @@ init() {
...
     trap "_shutdown" EXIT
 
+    if _should_skip_kernel_module_reload; then
+        echo "The NVIDIA driver is already loaded with the desired configuration, performing userspace-only install"
+        _unmount_rootfs
+        _install_userspace_components
+        _resolve_kernel_type || exit 1
+        _move_kernel_module_sources
+        _mount_rootfs
+        _start_daemons
```
cdesiniotis: Shouldn't the order here be reversed? For the full install path, we start the daemons before mounting the driver rootfs. Should we retain that order to be consistent?

karthikvetrivel: Updated

cdesiniotis (approved):

ubuntu22.04/nvidia-driver
```diff
@@ -658,35 +636,112 @@ _start_vgpu_topology_daemon() {
...
+        echo "Starting NVIDIA fabric manager daemon..."
+        nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
+    fi
+}
+
+_read_stored_digest() {
+    cat /run/nvidia/nvidia-driver.state 2>/dev/null || echo ""
+}
+
+_get_current_digest() {
```
shivakunv: the function is not used anywhere

ubuntu22.04/nvidia-driver
```diff
@@ -658,35 +636,112 @@ _start_vgpu_topology_daemon() {
...
+    local current_digest="${DRIVER_CONFIG_DIGEST:-}"
+    [ -z "${current_digest}" ] && return 1
+    local stored_digest=$(cat /run/nvidia/nvidia-driver.state 2>/dev/null || echo "")
+    [ "${current_digest}" = "${stored_digest}" ]
+}
+
+_store_driver_digest() {
+    local digest_file="${RUN_DIR}/nvidia-driver.state"
+    echo "Storing driver configuration digest..."
+    echo "${DRIVER_CONFIG_DIGEST}" > "$digest_file"
```
shivakunv: unbounded variable 
```
 local current_digest="${DRIVER_CONFIG_DIGEST:-}"
echo "${current_digest}" > "$digest_file"
```

shivakunv:
> @shivakunv Please note the changes added by @karthikvetrivel in this PR. These changes have to be added to the `rhel10` directory you're working on

Added and tested for RHEL10.
I have also added Krthik as a reviewer for PR: https://github.com/NVIDIA/gpu-driver-container/pull/496

ubuntu22.04/nvidia-driver
karthikvetrivel: Bounded variable.

ubuntu22.04/nvidia-driver
karthikvetrivel: Good catch, removed unused functions.

cdesiniotis:
🚀

PR #537, opened 2026-01-15
Sync sle15 dockerfile with features available in other distributions 

fcrozat:
Use official gitlab url to grab license rather than fetching a container with it.

Allow to build both proprietary and open drivers

update CA certificates data if additional pem are found

### DIFF
--- a/sle15/Dockerfile
+++ b/sle15/Dockerfile
@@ -1,42 +1,54 @@
 ARG SLES_VERSION
-FROM nvcr.io/nvidia/cuda:13.1.1-base-ubi8 as license
+ARG GOLANG_VERSION=1.25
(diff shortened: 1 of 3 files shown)


rajathagasthya:
@fcrozat Thanks for the contribution! Unfortunately, we don't plan to add support for new platforms other than the [officially supported platforms](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/platform-support.html#supported-operating-systems-and-kubernetes-platforms) currently. We've removed the unused SLE15 build scripts from his repo recently in #739. 

Our recommendation is to use build scripts in this repository as reference and maintain the driver images yourself. Hope this is helpful!

PR #538, opened 2026-01-15
Add Fabric Manager Shared NVSwitch virtualization model support

mresvanis:
This PR adds [Fabric Manager](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#) (FM) [Shared NVSwitch virtualization model](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#shared-nvswitch-virtualization-model) support when NVSwitch devices are detected and the newly introduced `FABRIC_MANAGER_FABRIC_MODE` env var is set to `1` (shared-nvswitch). 

No changes introduced when `FABRIC_MANAGER_FABRIC_MODE=0` (default FM mode - full-passthrough), which is the current flow when NVSwitch devices are detected.

Relates to: https://github.com/NVIDIA/gpu-operator/pull/2045

### Changes

- Add env var `FABRIC_MANAGER_FABRIC_MODE` to control fabric manager `FABRIC_MODE` (defaults to `0` for `full-passthrough`, `1` for `shared-nvswitch`).
- Update fabric manager config to the shared-nvswitch fabric mode.
- Configure UNIX socket communication instead of TCP.
- Create GPU physical module ID to PCIe address mapping JSON file via `nvidia-smi`.
- Do not start `nvidia-persistenced`  since GPU devices should be bound to `vfio-pci` by the `vfio-manager` in the next step.

### DIFF
--- a/rhel9/nvidia-driver
+++ b/rhel9/nvidia-driver
@@ -20,6 +20,7 @@ RHEL_MAJOR_VERSION=9
 RHEL_MINOR_VERSION=${RHEL_MINOR_VERSION:-""}
 KERNEL_MODULE_TYPE=${KERNEL_MODULE_TYPE:-auto}
 MODPROBE_CONFIG_DIR="/etc/modprobe.d"
+FABRIC_MANAGER_FABRIC_MODE=${FABRIC_MANAGER_FABRIC_MODE:-0}
 
 DRIVER_ARCH=${TARGETARCH/amd64/x86_64} && DRIVER_ARCH=${DRIVER_ARCH/arm64/aarch64}
 echo "DRIVER_ARCH is $DRIVER_ARCH"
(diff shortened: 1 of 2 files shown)


rhel9/nvidia-driver
```diff
@@ -715,25 +801,7 @@ _start_daemons() {
...
-                                               --fm-pid-file $fm_pid_file \
-                                               --nvlsm-config-file $nvlsm_config_file \
-                                               --nvlsm-pid-file $nvlsm_pid_file
-
-    # If not a NVLink5+ switch, check for the presence of NVLink4 (or below) switches
-    elif _assert_nvswitch_system; then
-        echo "Starting NVIDIA fabric manager daemon..."
-        nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
-    fi
+    _setup_fabric_manager "${fmpm_socket_path}"
```
cdesiniotis, line 805: I believe `fmpm_socket_path` will be unset here. This local variable is declared in a different function. (there was a refactor merged recently).

mresvanis: Thank you very much for catching that! It was indeed due to a rebase of this change for a refactor that was merged. I moved the `local fmpm_socket_path` variable definition in the `start_daemons()` function, as I think that that's the relevant part for this definition. Does that make sense?

cdesiniotis: Yes, makes sense. Resolving this comment.

cdesiniotis:
Closing in favor of https://github.com/NVIDIA/gpu-driver-container/pull/893

PR #563, opened 2026-01-19
Fix: Precompiled gh download Rate Limit Error

shivakunv:
Using `gh run download `in a loop with the paginate option results in multiple queries to download artifacts, which leads to a rate limit error. 

Implemented artifact download logic and removed the use of the `gh run` command from the pipeline.

### DIFF
--- a/.github/workflows/precompiled.yaml
+++ b/.github/workflows/precompiled.yaml
@@ -188,6 +188,13 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Download all kernel-version artifacts
+        uses: actions/download-artifact@v7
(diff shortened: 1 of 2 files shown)


.github/workflows/precompiled.yaml
```diff
@@ -188,6 +188,13 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Download all kernel-version artifacts
+        uses: actions/download-artifact@v4
```
tariq1890: why are you using an old version of this GH action?

shivakunv: modified

tariq1890 (approved):

PR #573, opened 2026-01-22
Add gnupg2 to CVE_UPDATES

JunAr7112:


tariq1890:
Please update the commit message and PR title as we are no longer updating `brotli`

PR #592, opened 2026-01-31
[fix] : update cache before running install

rahulait:
Fixes https://github.com/NVIDIA/gpu-operator/issues/1244

### Root cause
Package versions are rotated/updated on mirrors over time. The _install_driver() function in the nvidia-driver script was calling `apt-get install` without first running `apt-get update`. This meant it was using stale package lists that referenced package versions (like python3.10_3.10.12-1~22.04.7) that had been rotated out of the Ubuntu repositories in favor of newer/patched versions. This resulted in 404 errors when fetching those stale packages and install script exited with error.

### Fix
The fix proposes running `apt-get update` before running any `apt install`.

### DIFF
--- a/ubuntu22.04/nvidia-driver
+++ b/ubuntu22.04/nvidia-driver
@@ -39,7 +39,11 @@ _enable_fips_if_required() {
 _update_package_cache() {
     if [ "${PACKAGE_TAG:-}" != "builtin" ]; then
         echo "Updating the package cache..."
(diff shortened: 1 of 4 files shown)


cdesiniotis (approved):
Thanks @rahulait. I assume this was only affecting the precompiled Ubuntu containers since I see you only added a call to `update_package_cache` there?

PR #594, opened 2026-02-03
Set external release jobs to use Staging image construction 

JunAr7112:


Makefile
```diff
@@ -50,7 +50,7 @@ OUT_IMAGE_VERSION = $(OUT_VERSION)-$(DRIVER_TAG)
 endif
 
 OUT_DIST ?= $(DIST)
-OUT_IMAGE_TAG = $(OUT_IMAGE_VERSION)-$(OUT_DIST)
+OUT_IMAGE_TAG = $(OUT_IMAGE_VERSION)-$(DIST)
```
tariq1890: we should not be dropping the use of `$OUT_DIST` here. It is needed for for the rhel driver image tags

.common-ci.yml
```diff
@@ -190,7 +184,7 @@ trigger-pipeline:
     # Since OUT_IMAGE_NAME and OUT_VERSION are set, this will push the CI image to the target
     - if [[ "${PRECOMPILED}" == "true" ]]; then ./scripts/precompiled.sh push; else make push-${DIST}-${DRIVER_VERSION}; fi
     - mkdir -p build-info
-    - echo "${CI_COMMIT_SHORT_SHA}-${DRIVER_VERSION}-${DIST} ${DRIVER_VERSION}-${OUT_DIST}" > build-info/${CI_PIPELINE_ID}-${DRIVER_VERSION}-${OUT_DIST}.txt
+    - echo "${VERSION}-${DRIVER_VERSION}-${DIST} ${DRIVER_VERSION}-${OUT_DIST}" > build-info/${CI_PIPELINE_ID}-${DRIVER_VERSION}-${OUT_DIST}.txt
```
tariq1890: Is this change necessary? Isn't VERSION already set to CI_COMMIT_SHORT_SHA?

Makefile
JunAr7112: Earlier, when we publishing we would just use the rhel8/rhel9 tags in Staging jobs with tags. This change would still use the OUT_DIST for building the publishing template. Do we also want to use the OUT_DIST for the Staging?

I changed it to be more consistent with the practice in the other projects + the pre publishing changes format.

.common-ci.yml
JunAr7112: This change is not strictly needed, but makes testing easier with the NSPECT job (in case we are building an image with a different VERSION than CI_COMMIT_SHORT_SHA

Makefile
JunAr7112: I updated this to continue using the OUT_DIST. We will instead use the OUT_DIST for the source_version. I think this format might be simpler than the initial changes

.common-ci.yml
```diff
@@ -307,9 +306,6 @@ trigger-pipeline:
     - .release
   rules:
     - if: $CI_COMMIT_TAG
-      variables:
```
tariq1890, line 310: Why do we need to remove this?

JunAr7112: I removed it to align with the Makefile rules we are aiming for:

ifeq ($(OUT_VERSION),)
OUT_IMAGE_VERSION = (DRIVER_TAG)
else
OUT_IMAGE_VERSION = (OUT_VERSION)-(DRIVER_TAG)
endif

OUT_DIST ?= (DIST)
OUT_IMAGE_TAG = (OUT_IMAGE_VERSION)-(OUT_DIST)
OUT_IMAGE = (OUT_IMAGE_NAME):(OUT_IMAGE_TAG)

That way we will push images of form (CI_COMMIT_SHORT_SHA)-(TAG)-(OUT_DIST)

rajathagasthya (approved):
This looks okay to me. I'd like someone else to also approve this before merging.

.common-ci.yml
```diff
@@ -194,7 +193,7 @@ trigger-pipeline:
     # Since OUT_IMAGE_NAME and OUT_VERSION are set, this will push the CI image to the target
     - if [[ "${PRECOMPILED}" == "true" ]]; then ./scripts/precompiled.sh push; else make push-${DIST}-${DRIVER_VERSION}; fi
     - mkdir -p build-info
-    - echo "${CI_COMMIT_SHORT_SHA}-${DRIVER_VERSION}-${DIST} ${DRIVER_VERSION}-${OUT_DIST}" > build-info/${CI_PIPELINE_ID}-${DRIVER_VERSION}-${OUT_DIST}.txt
+    - echo "${VERSION}-${DRIVER_VERSION}-${OUT_DIST} ${DRIVER_VERSION}-${OUT_DIST}" > build-info/${CI_PIPELINE_ID}-${DRIVER_VERSION}-${OUT_DIST}.txt
```
tariq1890: Is this change really needed? It seems to me that `${VERSION}` and `${CI_COMMIT_SHORT_SHA}` are already the same, so I'd prefer to keep it how it was initially. It's more intuitive. Same for `${DIST}` and `${OUT_DIST}`

tariq1890 (changes requested):

.common-ci.yml
tariq1890: Then this change is not exactly relevant to what the PR wants to achieve. I'd prefer if we continued referencing `CI_COMMIT_SHORT_SHA` like before

.common-ci.yml
rajathagasthya: I think it's because Makefile constructs the source image tag from `VERSION` and not `CI_COMMIT_SHORT_SHA` directly. They are the same in the default case, but `VERSION` [can be overridden ](https://github.com/NVIDIA/gpu-driver-container/blob/6684a8a1378d0a93d2c9c02727b873b2f246a6e1/.common-ci.yml#L315) and then it wouldn't match.

tariq1890: This line just populates the `build-info` file though.

JunAr7112: @tariq1890 I can revert VERSION to CI_COMMIT_SHORT_SHA if that's preferred. However, OUT_DIST would definitely need to be referenced directly - it might not necessarily be equivalent to DIST.

release:ngc-rhcos4.17:
  extends:
    - .release:ngc
    - .dist-rhel9
  variables:
    OUT_DIST: "rhcos4.17"

.common-ci.yml
```diff
@@ -175,8 +175,7 @@ trigger-pipeline:
     # Define the source image for the release
     IMAGE_NAME: "${CI_REGISTRY_IMAGE}"
     VERSION: "${CI_COMMIT_SHORT_SHA}"
-    # OUT_VERSION is overridden for external releases
-    OUT_VERSION: "${CI_COMMIT_SHORT_SHA}"
+    OUT_VERSION: "${VERSION}"
```
cdesiniotis: Is there a specific reason for this change? IIUC nothing effectively changes, but I could be missing something....

.common-ci.yml
```diff
@@ -194,7 +193,7 @@ trigger-pipeline:
     # Since OUT_IMAGE_NAME and OUT_VERSION are set, this will push the CI image to the target
     - if [[ "${PRECOMPILED}" == "true" ]]; then ./scripts/precompiled.sh push; else make push-${DIST}-${DRIVER_VERSION}; fi
     - mkdir -p build-info
-    - echo "${CI_COMMIT_SHORT_SHA}-${DRIVER_VERSION}-${DIST} ${DRIVER_VERSION}-${OUT_DIST}" > build-info/${CI_PIPELINE_ID}-${DRIVER_VERSION}-${OUT_DIST}.txt
+    - echo "${CI_COMMIT_SHORT_SHA}-${DRIVER_VERSION}-${OUT_DIST} ${DRIVER_VERSION}-${OUT_DIST}" > build-info/${CI_PIPELINE_ID}-${DRIVER_VERSION}-${OUT_DIST}.txt
```
cdesiniotis, line 196: Is this required? 

Note, lots of our `rhcos` / `rhel` based tags are identical digests. So is there any harm in having the content of the build info files look like the below?
```
<sha>-<driver-version>-rhel9 <sha>-<driver-version>-rhel9.6
<sha>-<driver-version>-rhel9 <sha>-<driver-version>-rhel9.7
<sha>-<driver-version>-rhel9 <sha>-<driver-version>-rhcos4.14
<sha>-<driver-version>-rhel9 <sha>-<driver-version>-rhcos4.15
<sha>-<driver-version>-rhel9 <sha>-<driver-version>-rhcos4.16
. . .
```

JunAr7112: The release:ngc jobs will push an image with the OUT_DIST variable set rather than DIST. The source_version is set to use the OUT_DIST in order to align with the image that gets pushed.

.common-ci.yml
JunAr7112: I can revert this I just checked and this shouldn't be needed

.common-ci.yml
JunAr7112: When discussing with Tariq earlier, we were thinking it would be best to keep on using the OUT_DIST for the release:ngc jobs.

cdesiniotis: Okay, thanks. Will resolve this comment then.

PR #602, opened 2026-02-07
[ubuntu] Support non-standard kernel version naming

### LINKED ISSUE #601: [Bug]: Ubuntu driver scripts assume standard kernel version naming, breaking custom kernels
**Describe the bug**

`_resolve_kernel_version()` in the Ubuntu nvidia-driver scripts assumes kernel versions follow Ubuntu's `major.minor.patch-revision-flavor` naming convention (e.g. `6.5.0-44-generic`). Custom kernels built for Ubuntu that use non-standard version naming (e.g. `6.5.13custom-00004-g00fabdb6286a`) are mangled into an incorrect version string, causing the driver build to fail because no matching `linux-headers-*` package exists.

The function extracts a version via regex from the `Version:` field and a flavor via `sed 's/[^a-z]*//'` from `uname -r`, then concatenates them. For non-standard kernels this produces a mangled, non-existent version. For example:

- `uname -r`: `6.5.13custom-00004-g00fabdb6286a`
- Version regex extracts: `6.5.13-00004`
- Flavor regex extracts: `custom-00004-g00fabdb6286a`
- Concatenated result: `6.5.13-00004-custom-00004-g00fabdb6286a` (does not exist)
- Correct value: `6.5.13custom-00004-g00fabdb6286a` (matches `uname -r` exactly)

**To Reproduce**

1. Run a kernel with a non-standard version string (e.g. a custom-built kernel using git-describe naming)
2. Ensure the corresponding `linux-headers-$(uname -r)` package is available in the apt repository
3. Run the gpu-driver-container. The driver build fails with "Could not resolve Linux kernel version"

**Expected behavior**

When `linux-headers-${KERNEL_VERSION}` already exists as a valid package for the exact `uname -r` output, the driver should use it as-is without attempting to decompose and reconstruct the version string.

**Environment (please provide the following information):**
 - `gpu-driver-container` source: HEAD of `main`
 - Host OS: Ubuntu 22.04
 - Kernel Version: Any non-standard naming (e.g. `6.5.13custom-00004-g00fabdb6286a`)
 - CPU Architecture: `x86_64`

**Information to attach**

Affected files, all Ubuntu/Debian variants share the same `_resolve_kernel_version()` implementation:
- `ubuntu16.04/nvidia-driver`
- `ubuntu18.04/nvidia-driver`
- `ubuntu20.04/nvidia-driver`
- `ubuntu22.04/nvidia-driver`
- `ubuntu24.04/nvidia-driver`
- `vgpu-manager/ubuntu20.04/nvidia-driver`
- `vgpu-manager/ubuntu22.04/nvidia-driver`
- `vgpu-manager/ubuntu24.04/nvidia-driver`

jfernandez:
Query apt-cache for linux-headers package info once and attempt to parse the version using the existing major.minor.patch-revision regex. When the regex does not match (e.g. custom/distro kernels with non-standard naming), keep KERNEL_VERSION unchanged instead of failing. This preserves the existing resolution logic for standard Ubuntu kernels while allowing custom kernels to work without modification.

Applied to all Ubuntu/Debian variants and vgpu-manager Ubuntu variants.

Closes: https://github.com/NVIDIA/gpu-driver-container/issues/601

### DIFF
--- a/ubuntu22.04/nvidia-driver
+++ b/ubuntu22.04/nvidia-driver
@@ -62,16 +62,25 @@ _update_ca_certificates() {
 
 # Resolve the kernel version to the form major.minor.patch-revision-flavor where flavor defaults to generic.
 _resolve_kernel_version() {
-    local version=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null | \
-      sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)
-    local kernel_flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//')
-    kernel_flavor="${kernel_flavor//virtual/generic}"
-
     echo "Resolving Linux kernel version..."
-    if [ -z "${version}" ]; then
+
+    local version_info=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null || true)
+    if [ -z "${version_info}" ]; then
         echo "Could not resolve Linux kernel version" >&2
         return 1
     fi
+
+    local version=$(printf "%s\n" "${version_info}" | \
+      sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+)\..*/\1-\3/p' | head -1)
+
+    if [ -z "${version}" ]; then
+        # Custom/non-standard kernel: keep KERNEL_VERSION unchanged.
+        echo "Proceeding with Linux kernel version ${KERNEL_VERSION}"
+        return 0
+    fi
+
+    local kernel_flavor=$(echo "${KERNEL_VERSION}" | sed 's/[^a-z]*//')
+    kernel_flavor="${kernel_flavor//virtual/generic}"
 
     KERNEL_VERSION="${version}-${kernel_flavor}"
     echo "Proceeding with Linux kernel version ${KERNEL_VERSION}"
--- a/ubuntu16.04/nvidia-driver
+++ b/ubuntu16.04/nvidia-driver
@@ -26,16 +26,25 @@ _cleanup_package_cache() {
 
 # Resolve the kernel version to the form major.minor.patch-revision-flavor where flavor defaults to generic.
 _resolve_kernel_version() {
-    local version=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null | \
-      sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)
-    local kernel_flavor=$(echo ${KERNEL_VERSION} | sed 's/[^a-z]*//')
-    kernel_flavor="${kernel_flavor//virtual/generic}"
-
     echo "Resolving Linux kernel version..."
-    if [ -z "${version}" ]; then
+
+    local version_info=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null || true)
+    if [ -z "${version_info}" ]; then
         echo "Could not resolve Linux kernel version" >&2
         return 1
     fi
+
+    local version=$(printf "%s\n" "${version_info}" | \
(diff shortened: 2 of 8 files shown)


ubuntu22.04/nvidia-driver
```diff
@@ -62,12 +62,21 @@ _update_ca_certificates() {
 
 # Resolve the kernel version to the form major.minor.patch-revision-flavor where flavor defaults to generic.
 _resolve_kernel_version() {
+    echo "Resolving Linux kernel version..."
+
+    # If headers for the exact running kernel exist, use KERNEL_VERSION as-is.
+    # This supports non-standard kernel naming (e.g. custom/distro kernels)
+    # where the upstream version regex would fail.
+    if apt-cache show "linux-headers-${KERNEL_VERSION}" > /dev/null 2>&1; then
```
shivakunv: `apt-cache `is queried twice. 
what we can do :

shivakunv:
```
   version_info=$(apt-cache show "linux-headers-${KERNEL_VERSION}" 2> /dev/null || true)
    if [ -z "${version_info}" ]; then
        echo "Could not resolve Linux kernel version" >&2
        return 1
    fi

     local version=$(printf "%s\n" "${version_info}" | \
      sed -nE 's/^Version:\s+(([0-9]+\.){2}[0-9]+)[-.]([0-9]+).*/\1-\3/p' | head -1)

    if [ -z "${version}" ]; then
        # custom kernel : keep KERNEL_VERSION unchanged.
        echo "Proceeding with Linux kernel version ${KERNEL_VERSION}"
        return 0
    fi
   ....... 
```

jfernandez: Applied

shivakunv:
@jfernandez Please share the test logs for both types of kernels, if available.

@rajathagasthya  LGTM. PTAL

jfernandez:
@shivakunv Testing surfaced a gap in the initial patch: the regex `([0-9]+).*` was accidentally matching our kernel's package version (`6.5.13-00004-g00fabdb6286a-1`), extracting `6.5.13-00004` and producing a mangled `KERNEL_VERSION` whose headers package doesn't exist, so the fallback never triggered. I pushed a fix changing `.*` to `\..* ` to require a dot after the revision, which is always present in standard Ubuntu versions but not in our custom format.

Logic is working as expected after this change, here are the logs for both our custom kernel and standard Ubuntu:

Custom kernel (Ubuntu 22.04, `6.5.13netflix-00004-g00fabdb6286a`):
```
Resolving Linux kernel version...
Proceeding with Linux kernel version 6.5.13netflix-00004-g00fabdb6286a
```

Standard Ubuntu kernel (Ubuntu 22.04 VM, `5.15.0-173-generic`):
```
Resolving Linux kernel version...
Proceeding with Linux kernel version 5.15.0-173-generic
```

shivakunv:
> @shivakunv Testing surfaced a gap in the initial patch: the regex `([0-9]+).*` was accidentally matching our kernel's package version (`6.x.y-00004-g00fabdb6286a-1`), extracting `6.x.y-00004` and producing a mangled `KERNEL_VERSION` whose headers package doesn't exist, so the fallback never triggered. I pushed a fix changing `.*` to `\..* ` to require a dot after the revision, which is always present in standard Ubuntu versions but not in our custom format.
> 
> Logic is working as expected after this change, here are the logs for both our custom kernel and standard Ubuntu:
> 
> Custom kernel (Ubuntu 22.04, `6.x.ynetflix-00004-g00fabdb6286a`):
> 
> ```
> Resolving Linux kernel version...
> Proceeding with Linux kernel version 6.x.ynetflix-00004-g00fabdb6286a
> ```
> 
> Standard Ubuntu kernel (Ubuntu 22.04 VM, `5.15.0-173-generic`):
> 
> ```
> Resolving Linux kernel version...
> Proceeding with Linux kernel version 5.15.0-173-generic
> ```
Thanks @jfernandez . 

Could you please provide the log files for both kerbel version ?

rahulait (approved):
Overall, LGTM. We only support ubuntu22.04 and higher, so changes to versions lower than that will not be used.

tariq1890 (approved):

rahulait:
Merging this in. These changes will be available once we release driver versions 535.309.01, 580.159.03 and 595.71.05 and onwards.

PR #566, opened 2026-01-21
Add SUBDIR mapping for vgpu rhel9.x targets

mvalsecchi-nv:
Close #494 .

On main, rhel9.6 targets resolve to the wrong subdirectory (`rhel9.6/` instead of `rhel9/`), as reported in #486.

Steps from [cloud-native docs](https://docs.nvidia.com/datacenter/cloud-native/openshift/latest/openshift-virtualization.html#nvidia-gpu-operator-with-openshift-virtualization).

Before the fix (on main):
```
 $ git log -1 --pretty=oneline
b36127925fece95b9e2edcc55d0dcd8e42636cdb (HEAD -> main, origin/main, origin/HEAD) Merge pull request #564 from NVIDIA/dependabot/github_actions/renovatebot/github-action-44.2.5
 $ export PRIVATE_REGISTRY=my.private.registry.example.com VGPU_HOST_DRIVER_VERSION=570.211.01 OS_TAG=rhel9.6
 $ GPU_HOST_DRIVER_VERSION=${VGPU_HOST_DRIVER_VERSION} IMAGE_NAME=${PRIVATE_REGISTRY}/vgpu-manager make build-vgpuhost-${OS_TAG}
DOCKER_BUILDKIT=1 \
        docker  build --pull \
                        --output=type=image,push=false \
                        --platform=linux/amd64 \
                        --tag my.private.registry.example.com/vgpu-manager:570.211.01-rhel9.6 \
                        --build-arg DRIVER_BRANCH="570" \
                        --build-arg DRIVER_VERSION="570.211.01" \
                        --build-arg GOLANG_VERSION="1.25.6" \
                        --build-arg CVE_UPDATES="" \
                        --build-arg CUDA_VERSION="" \
                         \
                        --file /home/user/gpu-driver-container/vgpu-manager/rhel9.6/Dockerfile \
                        /home/user/gpu-driver-container/vgpu-manager/rhel9.6
[+] Building 0.0s (0/0)                                                                                                                                                                        docker:default
ERROR: failed to build: unable to prepare context: path "/home/user/gpu-driver-container/vgpu-manager/rhel9.6" not found
make: *** [Makefile:290: build-vgpuhost-rhel9.6] Error 1
```

After the fix:
```
$ export PRIVATE_REGISTRY=my.private.registry.example.com VGPU_HOST_DRIVER_VERSION=570.211.01 OS_TAG=rhel9.6
[...]

PR #610, opened 2026-02-13
Updates to vGPU Manager build

mattwittwer:
Fix vGPU Manager build for OCP 4.19+ and pciutils availability in air-gapped environments

The documented OS_TAG=rhel9.6 for OpenShift 4.19+ fails because the Makefile SUBDIR overrides only match the "rhcos" prefix, not "rhel9". Add SUBDIR mappings for rhel9-prefixed targets across driver, vgpu-guest, and vgpu-host build targets.

The ocp_dtk_entrypoint script copies the lspci binary to the shared directory for the DTK container, but does not copy its library dependency (libpci.so) or the setpci binary also required by sriov-manage. In air-gapped environments where the dnf fallback cannot reach cdn-ubi.redhat.com, this causes VF creation to fail silently, preventing vGPU mediated devices from being created.

Copy libpci.so* and setpci alongside lspci into the shared directory, and export LD_LIBRARY_PATH in the DTK container so the binaries can find their libraries.

### DIFF
--- a/vgpu-manager/rhel9/ocp_dtk_entrypoint
+++ b/vgpu-manager/rhel9/ocp_dtk_entrypoint
@@ -32,8 +32,14 @@ nv-ctr-run-with-dtk() {
            /usr/local/bin/ocp_dtk_entrypoint \
            /usr/local/bin/nvidia-driver \
            /usr/sbin/lspci \
+           /usr/sbin/setpci \
            /driver \
            "$DRIVER_TOOLKIT_SHARED_DIR/"
+
+        # Copy pciutils library dependencies for the DTK container
+        # (the DTK image does not have libpci, and dnf fallback fails in air-gapped networks)
+        mkdir -p "$DRIVER_TOOLKIT_SHARED_DIR/lib64"
+        cp /usr/lib64/libpci.so* "$DRIVER_TOOLKIT_SHARED_DIR/lib64/"
 
         env | sed 's/=/="/' | sed 's/$/"/' > "$DRIVER_TOOLKIT_SHARED_DIR/env"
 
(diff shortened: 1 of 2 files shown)


vgpu-manager/rhel9/ocp_dtk_entrypoint
```diff
@@ -32,9 +32,15 @@ nv-ctr-run-with-dtk() {
...
            /usr/local/bin/nvidia-driver \
            /usr/sbin/lspci \
+           /usr/sbin/setpci \
            /driver \
            "$DRIVER_TOOLKIT_SHARED_DIR/"
 
+        # Copy pciutils library dependencies for the DTK container
+        # (the DTK image does not have libpci, and dnf fallback fails in air-gapped networks)
+        mkdir -p "$DRIVER_TOOLKIT_SHARED_DIR/lib64"
+        cp /usr/lib64/libpci.so* "$DRIVER_TOOLKIT_SHARED_DIR/lib64/"
```
cdesiniotis, line 42: Question -- do we know if libpci.so will always be installed at `/usr/lib64`?

mattwittwer: Specifically for RHEL9 this is the standard library path and should be the same. But there may be a better approach to find this dynamically for compatibility against other OS versions or architectures. 

This change just allowed us to make progress in accessing libpci within the vgpu-manager image at runtime.

karthikvetrivel:
Thanks for this, @mattwittwer — appreciate you digging into both issues.

  Since this PR was opened, the changes here have largely landed via other paths:

  - The vgpuguest/vgpuhost Makefile mappings were merged in #566.
  - The air-gap lspci/setpci/libpci fix was merged in #618, addressing issue #616. That version also covers rhel8, uses lib/ instead of lib64/ to avoid the arch-specific path @cdesiniotis flagged, and keeps the bin/lib layout inside dtk-build-driver().

  The only piece here not already on main is the build-rhel9%: SUBDIR = rhel9 mapping for the standard driver target. If
   that path is actually exercised (the OpenShift docs direct users to rhcos* tags for the non-vGPU driver), it would be
   worth a small follow-up PR; otherwise I think we can close this one out.

  Closing in favor of #566 and #618 — but let me know if I'm missing something and happy to reopen.

PR #621, opened 2026-02-25
align package install code for regular and pre-compiled ubuntu drivers

rahulait:
This change aligns the package install code for both regular and pre-compiled drivers for ubuntu.


rahulait:
Tested ubuntu-22.04 precompiled image with following values:
```
driver:
  enabled: true
  nvidiaDriverCRD:
    enabled: false
    deployDefaultCR: false
  kernelModuleType: open
  repository: ghcr.io/nvidia
  image: driver
  usePrecompiled: true
  version: 1ae29e79-580
  imagePullPolicy: Always
  ```
  All operands came up fine. Example:
  ```
  root@ipp2-1669:~# kgpo | grep driver
nvidia-driver-daemonset-5.15.0-171-generic-ubuntu22.04-hv26g   1/1     Running     0          3m41s
root@ipp2-1669:~#
```

PR #636, opened 2026-03-04
allow specific precompiled images to be rebuilt

rahulait:
This PR adds the following capability:

Manual workflow inputs for targeted builds — each matrix dimension (driver branch, kernel flavor, dist, LTS kernel) can be filtered to a single value via workflow_dispatch inputs, instead of always building the full matrix.

Force rebuild flag — a new force_rebuild input that bypasses the ghcr.io tag-existence check (nvcr.io is still treated as authoritative). This will enable us to build and push updated images to ghcr.io which then can be published to nvcr.io

### DIFF
--- a/.github/workflows/precompiled.yaml
+++ b/.github/workflows/precompiled.yaml
@@ -19,6 +19,28 @@ on:
   schedule:
     - cron: '00 09 * * *'
   workflow_dispatch:
+    inputs:
+      driver_branch:
+        description: 'Driver branch to build (e.g. 535). Leave empty for all.'
+        required: false
+        default: ''
+      kernel_flavor:
+        description: 'Kernel flavor to build (e.g. azure). Leave empty for all.'
+        required: false
+        default: ''
+      dist:
+        description: 'Distribution to build (e.g. ubuntu22.04). Leave empty for all.'
+        required: false
+        default: ''
+      lts_kernel:
+        description: 'LTS kernel series to build (e.g. 6.8). Leave empty for all.'
+        required: false
+        default: ''
+      force_rebuild:
+        description: 'Force rebuild and re-publish even if the image tag already exists'
+        required: false
+        type: boolean
+        default: false
 
 concurrency:
   group: precompiled-driver-build-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


rahulait:
> To override image in staging , we need to change:
> 
> ```
> precompiled.sh :- 
> 
> function imageExists(){
> 	regctl manifest get $1 --list > /dev/null && return 0 || return 1
> } 
> ```
> 
> We should add an md5sum check between the staging and ghcr repo to verify whether the image is same or not .

Fixed this in change https://github.com/NVIDIA/gpu-driver-container/compare/dc1ee9b75cb65c00cb3a4c8a9bf15779e11e53e4..65f583ce429aa61f709a5194346e18009afda582

.github/workflows/precompiled.yaml
```diff
@@ -213,7 +252,12 @@ jobs:
           done))
           fi
           source ./tests/scripts/ci-precompiled-helpers.sh
-          KERNEL_VERSIONS=($(get_kernel_versions_to_test KERNEL_FLAVORS[@] DRIVER_BRANCHES[@] $DIST $LTS_KERNEL))
+          KERNEL_VERSIONS=($(get_kernel_versions_to_test KERNEL_FLAVORS[@] DRIVER_BRANCHES[@] $DIST $LTS_KERNEL $FORCE_REBUILD))
+          rc=$?
```
shivakunv, line 261: I believe we can skip the network/registry connectivity error check, as I haven't encountered this issue.
Reason:
1. We didn't face this problem in the earlier precompiled-build stage.
2. There were no issues during registry login.
3. Therefore, we shouldn't expect this issue here. Omitting this error check will help simplify the code.

since we're already skipping e2e/publish if it happens, any issues will be resolved by the next day's scheduled job.

tests/scripts/findkernelversion.sh
```diff
@@ -32,12 +33,33 @@ if [ -n "$artifact" ]; then
...
+else
+    ghcr_tags=$(regctl tag ls ghcr.io/nvidia/driver 2>&1)
+    ghcr_status=$?
+    if [[ $ghcr_status -ne 0 ]]; then
+        echo "failed to list tags from ghcr.io/nvidia/driver (exit $ghcr_status): $ghcr_tags" >&2
+        export should_continue=false
+        export regctl_error=true
+        return 1
+    fi
+    if echo "$ghcr_tags" | grep -q "^${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST}$"; then
```
shivakunv, line 60: this check will be sufficient

.github/workflows/precompiled.yaml
rahulait: Say an image exists in ghcr and there is some issue connecting to ghcr and ngc (network issue, rate limited, etc) when regctl commands are run, so based on our current logic:
```
regctl tag ls  nvcr.io/nvidia/driver | grep "^${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST}$" || status_nvcr=$?
regctl tag ls  ghcr.io/nvidia/driver | grep "^${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST}$" || status_ghcr=$?
if [[ $status_nvcr -eq 0 || $status_ghcr -eq 0 ]]; then
    export should_continue=false
else
    export should_continue=true
fi
```

status_nvcr will be non-zero and status_ghcr will be non-zero as well. In such a case, it will set `export_should_continue=1` which is a false positive. This check here is just making sure we return with error code in such cases and log why it failed.

shivakunv (approved):
LGTM

rahulait:
Closing as not a priority atm.

Issue #347, opened 2025-06-13
apt(-get) operations that require network: do not suppress stdout/err, timeout-control, retry


jgehrcke:
I think that we were bitten by a lack of a meaningful TCP connect() timeout in `apt-get ...` when we recently saw the driver container hang in places like

```
Installing Linux kernel headers...
```
```
Installing Linux kernel module files...
```
The pod would hang there for a long time and eventually restart (after about ~20 minutes).

It took quite a bit of back and forth before we understood that this is probably networking-related and not filesystem-related. We would have understood that faster if we would have seen stdout of `apt-get`: its stdout would have revealed that it is _trying to_ fetch data from remote infra.

Looking at code, in many places we invoke `apt-get` with this pattern: 
```
apt-get -qq install ... > /dev/null
```

Changes that we should make:
- never suppress stdout (also not stderr)
- add an overall meaningful timeout around the command, and retry it (a good timeout constant could be: maximum expected time of execution on a slow system times 5)

A bit of background / big picture about this type of problem: I deeply care about debuggabililty. That often means that we want to support humans in their debugging efforts.

One of the not-so-much fun unnecessary time sinks during debugging is when there's no log output and something just hangs in an operation that
1) could have been logged (so that one knows where something hangs)
2) could have been timeout-controlled and retried (often leading to success, instead of hanging forever)

Rather often that relates to deliberately setting a TCP connect() timeout. For the record, I have frequently talked about and written about  the immense importance and power of combining i) timeout control, ii) retries, and iii) logging elsewhere, such as in
- https://github.com/conbench/conbench/issues/801
- https://github.com/python/cpython/issues/89953#issuecomment-1458261299
- https://github.com/astral-sh/uv/issues/8144

PR #643, opened 2026-03-09
apt(-get) operations that require network: do not suppress stdout/err…

shivakunv:
github issue : [https://github.com/NVIDIA/gpu-driver-container/issues/347](https://github.com/NVIDIA/gpu-driver-container/issues/347)

### DIFF
--- a/base/generate-ci-config
+++ b/base/generate-ci-config
@@ -13,7 +13,7 @@ SUPPORTED_KERNELS=""
 
 # Silent apt update
 export DEBIAN_FRONTEND=noninteractive
-apt-get update -y -qq > /dev/null
+apt-get update -y
 
 # Generate a list of all kernel versions which have NVIDIA precompiled driver packages available.
 SUPPORTED_KERNELS_LINUX_SIGNATURES_LIST=$(apt-cache search "linux-signatures-nvidia.*${LTS_KERNEL}.*${KERNEL_FLAVOR}" | awk '{print $1}'  | sed -e "s/^.*${LTS_KERNEL}/${LTS_KERNEL}/" | sort -n -t'-' -k2| grep "${KERNEL_FLAVOR}$")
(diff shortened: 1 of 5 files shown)


ubuntu22.04/precompiled/nvidia-driver
```diff
@@ -75,10 +75,24 @@ nvsdm_install() {
   fi
 }
 
+# Run apt-get -qq quietly, on failure print actual error to stdout and exit.
+_apt_quiet() {
```
rajathagasthya: Why are we running `apt-get` quietly if the motivation is to have verbose logs?

shivakunv: _verbose logs will be generated only when errors occur, will be ignored onsucces._

Also, go back to plain `apt-get -qq` ... with no `> /dev/null or 2>/dev/null.`
`-qq` only quiets apt-get itself. It does not silence dpkg or maintainer scripts .
So you still get output such as :

```
“(Reading database …)”
“Selecting previously unselected package …”
“Unpacking …” / “Setting up …”
“Processing triggers for …” (e.g. man-db, systemd)
```
We can expect a text output of about 10 to 100 lines(few kbs).

_apt_quiet function will suppress this.

rajathagasthya: As we discussed, let's drop `-qq` and redirection to `/dev/null`. More verbose logging is okay.

shivakunv:
@rajathagasthya PTAL

ubuntu22.04/precompiled/nvidia-driver
tariq1890: I agree with @rajathagasthya here. @shivakunv Please address Rajath's comment above

shivakunv: addressed it

rajathagasthya (approved):
LGTM

Can you post logs from this change? I'd like to see how verbose they are.

PR #683, opened 2026-04-03
Skip modprobe in peermem container when module is already loaded

karthikvetrivel:
**Problem**                                                                                                                 
                                                                                                                          
  When a driver pod is deleted and recreated, the fast-path optimization skips kernel module compilation/installation     
  since modules are already loaded in memory. However, the peermem sidecar unconditionally runs `chroot /run/nvidia/driver 
  modprobe nvidia-peermem`, which fails because `/lib/modules/<kernel>/` has no `.ko` files on the fast path.                  
                                                            
  **Fix**

  Add an early check for /sys/module/nvidia_peermem/refcnt in reload_nvidia_peermem() before attempting modprobe. This mirrors the pattern used in nvidia-validator's `isNvidiaModuleLoaded()`.
                                                                                                                          
  **Test plan**                                                 

  - Deleted a running driver pod (kubectl delete pod <driver-pod>)                                                              
  - Verified new pod reaches Ready state with all containers running
  - Verified k8s-driver-manager logs show "skipping the uninstallation"                                                     
  - Verified nvidia-peermem-ctr logs show "already loaded, skipping modprobe"

### DIFF
--- a/rhel8/nvidia-driver
+++ b/rhel8/nvidia-driver
@@ -905,6 +905,12 @@ reload_nvidia_peermem() {
     fi
     # get any parameters provided for nvidia-peermem
     _get_module_params && set +o nounset
+    # If nvidia-peermem is already loaded (e.g. fast-path restart), skip modprobe
+    if [ -f /sys/module/nvidia_peermem/refcnt ]; then
+        echo "nvidia-peermem module already loaded, skipping modprobe"
+        sleep inf
+        trap "echo 'Caught signal'; exit 1" HUP INT QUIT PIPE TERM
+    fi
     if chroot /run/nvidia/driver modprobe nvidia-peermem "${NVIDIA_PEERMEM_MODULE_PARAMS[@]}"; then
         if [ -f /sys/module/nvidia_peermem/refcnt ]; then
             echo "successfully loaded nvidia-peermem module, now waiting for signal"
--- a/rhel10/nvidia-driver
+++ b/rhel10/nvidia-driver
@@ -922,6 +922,12 @@ reload_nvidia_peermem() {
     fi
     # get any parameters provided for nvidia-peermem
     _get_module_params && set +o nounset
+    # If nvidia-peermem is already loaded (e.g. fast-path restart), skip modprobe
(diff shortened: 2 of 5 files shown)


rhel8/nvidia-driver
```diff
@@ -905,6 +905,12 @@ reload_nvidia_peermem() {
     fi
     # get any parameters provided for nvidia-peermem
     _get_module_params && set +o nounset
+    # If nvidia-peermem is already loaded (e.g. fast-path restart), skip modprobe
+    if [ -f /sys/module/nvidia_peermem/refcnt ]; then
```
tariq1890, line 909: What if we called this method instead - `_should_skip_kernel_module_reload` ?

Wouldn't that make it consistent with the main container?

karthikvetrivel: The digest environment variable isn't set on the peermem sidecar container. This isn't a big issue, I just wanted to flag that it would require a change in the operator. 

I also agree that `_should_skip_kernel_module_reload` is a good proxy for asking "are we on the fast path." In most cases this is fine — if we're running fast path, peermem should already be loaded into memory. However, in the case that peermem failed on a prior boot (unlikely) and then the pod restarted, the container will report success while peermem is missing. 

Let me know what you think @tariq1890

tariq1890: Note that the peermem module is first loaded by the main container. The peermem sidecar container performs reloads of peermem module as and when required

tariq1890: Resolving this comment as it is more thorough to check the refcnt of `nvidia_peermem` rather than a fast path check.

tariq1890 (approved):
Thanks @karthikvetrivel ! Let's get one more approve before merging

JunAr7112:
This LGTM.

JunAr7112 (approved):

cdesiniotis:
Do we also need to make this change in the `rhel10` directory?

karthikvetrivel:
> Do we also need to make this change in the `rhel10` directory?

Good catch @cdesiniotis, fixed.

PR #707, opened 2026-04-15
add support for 595 branch drivers for precompiled images

rahulait:
Release precompiled drivers for 595 branch as well.

The Ubuntu 22.04 precompiled image needs multiverse because NVIDIA userspace packages like nvidia-utils-595-server are published there. It also needs restricted for the Canonical-signed NVIDIA kernel module packages, so both components are required for the build.

### DIFF
--- a/.common-ci.yml
+++ b/.common-ci.yml
@@ -98,7 +98,7 @@ trigger-pipeline:
 .driver-versions-precompiled-ubuntu22.04:
   parallel:
     matrix:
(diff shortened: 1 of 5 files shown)


tariq1890 (approved):

tariq1890:
Should we remove `535` ?

rahulait:
> Should we remove `535` ?

I'll do that in a follow up PR. Thanks for reminding its EOL now.

PR #768, opened 2026-05-09
[vgpu-manager] support custom CA certificate injection at build time

Shivkumar13:
Add an opt-in CUSTOM_CA_CERTS_DIR build arg to all four vGPU Manager Dockerfiles (rhel8, rhel9, ubuntu22.04, ubuntu24.04) and wire it through the Makefile build-vgpuhost-% target. Users behind a TLS-intercepting corporate proxy can drop *.crt files into vgpu-manager/<os>/certs/ and rebuild without editing the Dockerfile. The directory is empty by default, so out-of-the-box builds are unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,6 +288,11 @@ push-vgpuguest-%: DIST = $(word 3,$(subst -, ,$@))
 # The vGPU host driver .run file is assumed to be present in the $SUBDIR/drivers/ directory.
 # VGPU_HOST_DRIVER_VERSION must be defined in the environment when invoking this target.
 VGPU_HOST_DRIVER_VERSION ?= ""
+# CUSTOM_CA_CERTS_DIR is the path (relative to the build context
+# vgpu-manager/$(SUBDIR)) of a directory containing custom CA certificates to
+# trust at build time. Default is "certs", an empty directory shipped in the
+# repo so the COPY in the Dockerfile is a no-op out of the box.
+CUSTOM_CA_CERTS_DIR ?= certs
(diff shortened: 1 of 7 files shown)


rahulait:
Thanks @Shivkumar13 . Lets get rid of ubuntu20.04 changes as we don't support it. Also, we should add support in ubuntu24.04 as well.

Is there a linked issue for this fix?

Shivkumar13:
@rahulait We don't have an issue, there is a customer case for this. I can open an issue or NVBug we want to track this.

rahulait:
Lets open internal bug for this then and you can add me to it. I would like to understand what the issue is and get others feedback as well.

rahulait:
@Shivkumar13 I was able to find the customer issue. Lets fix this PR by removing ubuntu20.04 change and adding support for ubuntu24.04 as well.

Shivkumar13:
@rahulait done with changes, PTAL.

rahulait (approved):
Thanks @Shivkumar13 , LGTM

tariq1890:
@Shivkumar13 @rahulait Do we need to make changes for these in `rhel10` too?

rahulait:
@tariq1890 rhel10 is not there under vgpu-manager. Maybe we should start supporting rhel10 as well?

tariq1890:
Yes, we should. Let's create a GH issue so we can track this.

Shivkumar13:
@tariq1890 @rahulait I have created an issue, please feel free to change the issue description if needed

https://github.com/NVIDIA/gpu-driver-container/issues/838

PR #777, opened 2026-05-12
base: replace Ubuntu CUDA base images with native Ubuntu base

shivakunv:
base/Dockerfile is used only by the precompiled image flow to run generate-ci-config and extract kernel version  at build time. 

Since this is only a compile-time requirement for determining the kernel version, switching the cuda Ubuntu base stages to native ubuntu images lets us cleanly remove the CUDA base image dependency .


rajathagasthya (approved):

base/Dockerfile
rahulait: Do we need this? If not, lets remove the ubuntu20.04 related code

rahulait: This might no longer be required as it won't be present.

rahulait: Lets optimize this file for new base image and remove code which is no longer needed as it won't be in the base image.

rahulait (approved):

PR #840, opened 2026-07-01
[renovate][ubuntu] update allowedVersions regex

tariq1890:
~I am not sure why @renovate bot isn't creating the PRs for `ubuntu24.04`, I've decided to create this PR to bump the base images for the time beings. Even a manual trigger of the @renovate workflow didn't seem to help.~

(Updated PR as per Rahul's suggestion)

This PR should make it possible for renovate to auto-bump the ubuntu24.04 base images to the latest tag (which has a `.`) - `noble-20260509.1`

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -26,7 +26,7 @@
             "matchFileNames": [
                 "ubuntu22.04/**"
             ],
-            "allowedVersions": "/^jammy(-\\d{8})?$/"
+            "allowedVersions": "/^jammy(-\\d{8}(\\.\\d+)?)?$/"
         },
         {
             "matchDatasources": [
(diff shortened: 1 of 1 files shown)


rahulait:
We shouldn't be doing manual changes and should fix renovate regex.

Current rule is `"allowedVersions": "/^noble(-\\d{8})?$/"` which limits the tag to not have dotted builds. The rule should be updated for both ubuntu22.04 and ubuntu24.04 to something like `"allowedVersions": "/^noble(-\\d{8}(\\.\\d+)?)?$/"` so that dotted versions are also allowed.

I would prefer us fixing renovate rule and then retriggering renovate to create the PR than manually doing it.

tariq1890:
I've updated the PR and renamed the commit. Please check now.

rahulait (approved):

PR #857, opened 2026-07-07
Add Ubuntu 26.04 driver container with package-based driver install

rajathagasthya:
Adds a new `ubuntu26.04/` driver container. For the passthrough (baremetal) driver type it
installs the NVIDIA driver from apt packages in the `ubuntu2604` CUDA repository and builds the
kernel modules with DKMS, replacing the `.run` installer. The vGPU driver type keeps the existing
`.run` flow.

## How it works

**Image build** installs the driver userspace (`nvidia-headless-no-dkms-open` plus the
display/codec libraries, matching what the `.run` file ships today) and the extra
packages (fabricmanager, nscq, imex, nvlsm, libnvsdm). It also downloads the driver metapackages,
the DKMS/kernel-source packages for both kernel module flavors, and `nvidia-firmware` (~185 MB
total) into a local `file:` apt repository baked into the image, then removes the CUDA network
repository so NVIDIA packages resolve only from the local one.

**Container start** installs the flavor-appropriate driver metapackage from the local
repository — `nvidia-open` for open, `cuda-drivers` for proprietary. At runtime, `apt`
skips everything already installed at build, so only the DKMS and kernel-source packages
are installed, and DKMS builds the modules against the host kernel headers.
`nvidia-firmware` is reinstalled at container start: the GPU operator mounts `/lib/firmware` in
the driver container from the host so the kernel can read the GSP firmware, and that mount
hides the copy installed at image build. The reinstall writes the files into the mounted host
directory, which the `.run` installer achieves by writing the firmware at install time. The fast path
mounts the driver rootfs before starting the daemons so that GPU initialization always finds the firmware.
`dkms autoinstall -k $(uname -r)` runs afterwards as a check: the DKMS package postinst
exits 0 even when it skips the build, so this rerun surfaces any build failure and fails init
instead of continuing without modules.

**Fast path** (driver already loaded, configuration unchanged) installs nothing: no apt, no
network. Runtime network access is needed only for kernel headers, same as the `.run` flow.

## Package landscape

The two NVIDIA-documented install entry points are `nvidia-open` (open) and `cuda-drivers`
(proprietary). Both are empty metapackage shims whose dependency trees share the userspace
and differ only in the kernel-module rows at the bottom:

```
nvidia-open (shim)                          cuda-drivers (shim)
└── nvidia-driver-open (shim)               └── nvidia-driver (shim)
    ├── libnvidia-compute  ← nvidia-smi, NVML, libcuda, MPS ──┐
    │     └── nvidia-persistenced                             │
    ├── libnvidia-gl (273 MB: GLX/EGL/Vulkan ICDs)            │  identical
    ├── libnvidia-decode / libnvidia-encode / libnvidia-fbc1  │  userspace
    ├── nvidia-kernel-common                                  │  on both
    │     ├── nvidia-firmware  ← GSP firmware                 │  sides
    │     └── nvidia-modprobe ────────────────────────────────┘
    │
    ├── nvidia-kernel-source-open   |   ├── nvidia-kernel-source (~88 MB)
    └── nvidia-dkms-open            |   └── nvidia-dkms
          └── dkms (make, patch, …) |         └── dkms
```

`nvidia-kernel-source[-open]` is the module source tree under `/usr/src`; `nvidia-dkms[-open]`
registers it with DKMS. The two flavors are mutually exclusive (`nvidia-kernel-source-open`
declares `Provides/Conflicts/Replaces: nvidia-kernel-source`). This makes the runtime
kernel module type change just an `apt-get install`.

The same trees split by install time: the left side carries the `open` tree's dependencies,
the right side contains the `proprietary` tree's. The packages that come via
`nvidia-headless-no-dkms-open` are boxed.

```
CONTAINER START — installed from the local file: repo, per KERNEL_MODULE_TYPE
═══════════════════════════════════════════════════════════════════════════════════
 open/auto: 3 debs, <100 KB                proprietary: 4 debs, ~89 MB

 nvidia-open (shim)                        cuda-drivers (shim)
 └── nvidia-driver-open (shim)             └── nvidia-driver (shim)
     ├── nvidia-dkms-open                      ├── nvidia-dkms
     │                                         ├── nvidia-kernel-source (~88 MB)
     │                                         │     evicts nvidia-kernel-source-open
     │                                         └──────────────────────────────────┐
═════╪════════════════════════════════════════════════════════════════════════════╪
     │  BUILD TIME — userspace_install, baked into the image rootfs (~1 GB)       │
     │   ┌─ via nvidia-headless-no-dkms-open (shim) ──────────────────────────┐   │
     ├───┼─ libnvidia-compute          ← nvidia-smi, NVML, libcuda, MPS       ┼───┤
     │   │    └── nvidia-persistenced                                         │   │
     ├───┼─ nvidia-kernel-common                                              ┼───┤
     │   │    ├── nvidia-firmware       ← GSP firmware                        │   │
     │   │    └── nvidia-modprobe                                             │   │
     ├───┼─ nvidia-kernel-source-open  ← open module source; open tree only   │   │
     │   └────────────────────────────────────────────────────────────────────┘   │
     │   ┌─ added explicitly (the headless shim omits them) ──────────────────┐   │
     ├───┼─ libnvidia-gl (273 MB)      ← GLX/EGL/Vulkan ICDs                  ┼───┤
     ├───┼─ libnvidia-decode / libnvidia-encode / libnvidia-fbc1              ┼───┤
     │   │   libnvidia-extra            (outside both metapackage trees)      │   │
     └───┼─ dkms                       ← via nvidia-dkms[-open]               ┼───┘
         └────────────────────────────────────────────────────────────────────┘
 left side = nvidia-driver-open (open)
 right side = nvidia-driver (proprietary)
```

All userspace packages are installed at build time. Container start adds only the metapackage
shims and the DKMS packages that build the kernel modules.

## Design decisions

1. **Install from packages, not the `.run` file (passthrough only).** Removes the
   `.run`/`nvidia-installer` dependency for the passthrough type.

2. **Driver metapackage plus a version pinning package, not a curated package list.** The build
   installs `nvidia-driver-pinning-<version>`, which pins the whole driver tree (driver
   metapackage, persistenced, fabricmanager, nscq, imex, libnvsdm) to the exact release, so builds
   are reproducible without hand-pinning each package. The metapackage is the NVIDIA-documented
   install path and needs no changes when package names change.

3. **Full driver, not headless.** `libnvidia-gl`, `libnvidia-decode`/`encode`/`fbc1` and the EGL
   libraries are kept for parity with the `.run` image. Graphics workloads consume them from the
   driver rootfs.

4. **`KERNEL_MODULE_TYPE=auto` resolves to open; no runtime detection.** Every driver branch in
   the `ubuntu2604` repository (>= 595) supports Turing+ GPUs only, where the open modules are the
   recommended default, so detection via `nvidia-installer --print-recommended-kernel-module-type`
   would always return "open". Proprietary remains an explicit opt-in (`cuda-drivers`).

5. **vGPU keeps the `.run` flow.** vGPU guest drivers need to be installed using `.run` files from
   the licensing portal with no package equivalent. Only the install step differs between the two
   types; module load, daemons, and the rootfs mount are shared.

6. **Driver payload baked into the image; no network dependency at container start.** Installing
   from the network CUDA repository at runtime would have made every container restart depend on
   repository access, which is a regression from the `.run` image where the fast path works offline.
   Cost: 0.84 GB compressed (amd64) vs 0.68 GB for the ubuntu24.04 `.run` image built from the
   same commit (arm64: 0.83 GB vs 0.60 GB).

## CI

- GitHub image build matrix includes ubuntu26.04, excluding the 580 driver (the `ubuntu2604`
  repository carries only branches >= 595).
- Renovate pins the base image to `resolute` tags.
- The QEMU emulator for arm64 cross-builds is now pinned per distribution. The previous
  `tonistiigi/binfmt:qemu-v6.2.0` pin (December 2021) cannot run the Ubuntu 26.04 arm64
  userland: the build hangs indefinitely, with no output, at the first apt layer under
  emulation. `qemu-v8.1.5` fixes that but segfaults in `libc-bin` on Ubuntu 22.04
  ([tonistiigi/binfmt#240](https://github.com/tonistiigi/binfmt/issues/240)), so no single
  version covers both. ubuntu22.04 keeps `qemu-v6.2.0`; every other distribution uses
  `qemu-v8.1.5` (all passed CI on it).

## Known limitations

- `init -m` / `MAX_THREADS` is not honored on the package path: the NVIDIA `dkms.conf` hardcodes
  `-j$(nproc)`, so DKMS's own `-j` option has no effect. The vGPU `.run` path still honors it via
  `--concurrency-level`.

## Testing

- [x] Local image build (amd64)
- [x] Container start on a GPU node: open flavor
- [x] Container start: proprietary flavor (`KERNEL_MODULE_TYPE=proprietary`)
- [x] Offline container start (no network): first boot and fast path
- [x] vGPU `.run` flow unchanged
- [x] arm64 (sbsa) build

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -33,6 +33,7 @@ jobs:
         dist:
           - ubuntu22.04
           - ubuntu24.04
+          - ubuntu26.04
           - rhel8
           - rhel9
           - rhel10
@@ -41,6 +42,9 @@ jobs:
           - rocky10
         ispr:
           - ${{github.event_name == 'pull_request'}}
+        exclude:
+          - dist: ubuntu26.04
+            driver: 580.173.02
       fail-fast: false
     steps:
       - uses: actions/checkout@v7
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -72,7 +72,11 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v4
         with:
-          image: tonistiigi/binfmt:qemu-v6.2.0
+          # The Ubuntu 26.04 arm64 userland needs a newer emulator than the qemu-v6.2.0
+          # used for every other distribution: v6.2.0 hangs on it, and on the
+          # resolute-20260707 userland v8.1.5 fails with ENOSYS and v9.2.2 with EINVAL
+          # (tar file extraction), so it is pinned to qemu-v10.1.3.
+          image: tonistiigi/binfmt:${{ matrix.dist == 'ubuntu26.04' && 'qemu-v10.1.3' || 'qemu-v6.2.0' }}
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
         with:
@@ -155,7 +159,11 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v4
         with:
-          image: tonistiigi/binfmt:qemu-v6.2.0
+          # The Ubuntu 26.04 arm64 userland needs a newer emulator than the qemu-v6.2.0
+          # used for every other distribution: v6.2.0 hangs on it, and on the
+          # resolute-20260707 userland v8.1.5 fails with ENOSYS and v9.2.2 with EINVAL
+          # (tar file extraction), so it is pinned to qemu-v10.1.3.
+          image: tonistiigi/binfmt:${{ matrix.dist == 'ubuntu26.04' && 'qemu-v10.1.3' || 'qemu-v6.2.0' }}
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
         with:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -59,6 +59,14 @@ include:
     - if: $CI_PIPELINE_SOURCE != "schedule"
 
 # Define the image build targets
+.image-build-ubuntu26.04:
+  extends:
+    - .driver-versions-ubuntu26.04
+    - .image-build-generic
+  rules:
+    - if: $CI_PIPELINE_SOURCE != "schedule"
+
+# Define the image build targets
 .image-build-rhel9:
   # Perform for each DRIVER_VERSION
   extends:
@@ -85,6 +93,11 @@ image-ubuntu24.04:
   extends:
     - .image-build-ubuntu24.04
     - .dist-ubuntu24.04
+
+image-ubuntu26.04:
+  extends:
+    - .image-build-ubuntu26.04
+    - .dist-ubuntu26.04
 
 image-rhel8:
   extends:
--- a/.nvidia-ci.yml
+++ b/.nvidia-ci.yml
@@ -88,6 +88,15 @@ variables:
       when: never
     - !reference [.image-pull-rules, rules]
 
+.image-pull-ubuntu26.04:
+  extends:
+    - .driver-versions-ubuntu26.04
+    - .image-pull-generic
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "schedule"
+      when: never
+    - !reference [.image-pull-rules, rules]
+
 .image-pull-rhel10:
   # Perform for each DRIVER_VERSION
   extends:
@@ -158,6 +167,11 @@ image-ubuntu24.04:
   extends:
     - .image-pull-ubuntu24.04
     - .dist-ubuntu24.04
+
+image-ubuntu26.04:
+  extends:
+    - .image-pull-ubuntu26.04
+    - .dist-ubuntu26.04
 
 image-rhel8:
   extends:
@@ -271,6 +285,17 @@ image-rocky10:
     - if: $CI_PIPELINE_SOURCE == "merge_request_event"
     - !reference [.pipeline-trigger-rules, rules]
 
+.scan-ubuntu26.04:
+  extends:
+    - .driver-versions-ubuntu26.04
+    - .scan-generic
+  rules:
+    - !reference [.scan-rules-common, rules]
+    - if: $CI_PIPELINE_SOURCE == "schedule"
+      when: never
+    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
+    - !reference [.pipeline-trigger-rules, rules]
+
 .scan-rhel10:
   # Repeat for each DRIVER_VERSION
   extends:
@@ -350,6 +375,22 @@ scan-ubuntu24.04-arm64:
     - .platform-arm64
   needs:
     - image-ubuntu24.04
+
+scan-ubuntu26.04-amd64:
+  extends:
+    - .scan-ubuntu26.04
+    - .dist-ubuntu26.04
+    - .platform-amd64
+  needs:
+    - image-ubuntu26.04
+
+scan-ubuntu26.04-arm64:
+  extends:
+    - .scan-ubuntu26.04
+    - .dist-ubuntu26.04
+    - .platform-arm64
+  needs:
+    - image-ubuntu26.04
 
 scan-precompiled-ubuntu24.04-amd64:
   variables:
@@ -507,6 +548,12 @@ release:ngc-ubuntu24.04:
     - .release:ngc
     - .dist-ubuntu24.04
     - .driver-versions-ubuntu24.04
+
+release:ngc-ubuntu26.04:
+  extends:
+    - .release:ngc
+    - .dist-ubuntu26.04
+    - .driver-versions-ubuntu26.04
 
 release:ngc-precompiled-ubuntu24.04:
   variables:
--- a/ubuntu26.04/Dockerfile
+++ b/ubuntu26.04/Dockerfile
@@ -1,0 +1,100 @@
+ARG BASE_IMAGE=ubuntu:resolute-20260707
+
+FROM ${BASE_IMAGE} AS build
+
+ARG TARGETARCH
+ARG GOLANG_VERSION
+
+# Arg to indicate if driver type is either of passthrough(baremetal) or vgpu
+ARG DRIVER_TYPE=passthrough
+ENV DRIVER_TYPE=$DRIVER_TYPE
+
+SHELL ["/bin/bash", "-c"]
+
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        apt-utils \
+        build-essential \
+        ca-certificates \
+        curl \
+        git  \
+        wget && \
+    rm -rf /var/lib/apt/lists/*
+
+# download appropriate binary based on the target architecture for multi-arch builds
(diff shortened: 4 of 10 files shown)


ubuntu26.04/nvidia-driver
```diff
@@ -0,0 +1,939 @@
...
+    #    --skip-module-load
+    #         Skip the test load of the NVIDIA kernel modules after the modules are built,
+    #         and skip loading them after installation is complete.
+    #
+    # Without this flag, a subtle bug can occur if the nvidia-installer fails to unload
+    # the NVIDIA kernel modules after the test load. The modules will remain loaded and
+    # any custom NVIDIA module parameters configured as input to the driver container
+    # will not be applied.
+    #
+    if [ "${DRIVER_BRANCH}" -ge "550" ]; then
```
tariq1890: We don't need this if condition, yes?

ubuntu26.04/nvidia-driver
```diff
@@ -0,0 +1,939 @@
...
+        install_args+=("--skip-module-load")
```
tariq1890: This is context for the `--skip-module-load` option.

>     Specify the --skip-module-load flag for versions of the nvidia-installer that support it. 
From the nvidia-installer help output:
>  --skip-module-load
      Skip the test load of the NVIDIA kernel modules after the modules are built,
      and skip loading them after installation is complete.
    Without this flag, a subtle bug can occur if the nvidia-installer fails to unload
    the NVIDIA kernel modules after the test load. The modules will remain loaded and
    any custom NVIDIA module parameters configured as input to the driver container
    will not be applied.
    
Do we need to do anything to achieve a similar behaviour with the driver packages?

ubuntu26.04/Dockerfile
```diff
@@ -0,0 +1,100 @@
+ARG BASE_IMAGE=ubuntu:resolute-20260627
```
tariq1890:
```suggestion
ARG BASE_IMAGE=ubuntu:resolute-20260707
```

ubuntu26.04/nvidia-driver
```diff
@@ -0,0 +1,939 @@
...
+
+    echo "Installing Linux kernel headers..."
+    apt-get install -y --no-install-recommends linux-headers-${KERNEL_VERSION}
+
+    echo "Installing Linux kernel module files..."
+    apt-get download linux-image-${KERNEL_VERSION} && dpkg -x linux-image*.deb .
+    { apt-get download linux-modules-${KERNEL_VERSION} && dpkg -x linux-modules*.deb . || true; }
+    # Ubuntu 26.04 kernel debs ship their payload under usr/lib/modules (merged-usr paths),
+    # unlike the lib/modules payload on 24.04 and earlier.
+    mv usr/lib/modules/${KERNEL_VERSION}/modules.* /lib/modules/${KERNEL_VERSION}
```
tariq1890, line 128: no leading slashes needed?

rajathagasthya: No, same as `ubuntu24.04`. This is a relative path inside `tmp_dir` above and refers to the extracted modules from `dpkg -x` in the previous line, which preserves the original directory layout after extraction. Only difference is the layout was `lib/modules/...` in Ubuntu 24.04, now it's `usr/lib/modules/...`

ubuntu26.04/nvidia-driver
rajathagasthya: Makes sense, removed.

ubuntu26.04/nvidia-driver
rajathagasthya: I had Claude inspect the packages to make sure it doesn't try to load modules after install. It confirmed the `nvidia-dkms-open` package's `postinst` only runs `update-initramfs` and does a `dkms` build + install. And the `dkms autoinstall` step in L502 only builds, installs and runs `depmod` (doesn't load anything via `modprobe`). The first load is from `_load_driver`, which runs after `_create_module_params_conf` has already written module params. 

So I don't think we need to do anything here.

tariq1890: Thanks, this is good to know. FYI @cdesiniotis

.github/workflows/image.yaml
```diff
@@ -41,6 +42,10 @@ jobs:
           - rocky10
         ispr:
           - ${{github.event_name == 'pull_request'}}
+        # The ubuntu2604 CUDA repository carries only driver branches >= 595
```
tariq1890: nit: this comment is not needed

.gitlab-ci.yml
```diff
@@ -58,6 +58,15 @@ include:
   rules:
     - if: $CI_PIPELINE_SOURCE != "schedule"
 
+# Define the image build targets
+.image-build-ubuntu26.04:
+  # Perform for each DRIVER_VERSION
```
tariq1890: do we really need this comment?

.nvidia-ci.yml
```diff
@@ -88,6 +88,16 @@ variables:
       when: never
     - !reference [.image-pull-rules, rules]
 
+.image-pull-ubuntu26.04:
+  # Perform for each DRIVER_VERSION
```
tariq1890: same as previous

.nvidia-ci.yml
```diff
@@ -271,6 +286,18 @@ image-rocky10:
     - if: $CI_PIPELINE_SOURCE == "merge_request_event"
     - !reference [.pipeline-trigger-rules, rules]
 
+.scan-ubuntu26.04:
+  # Repeat for each DRIVER_VERSION
```
tariq1890: same as previous

versions.mk
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 # DRIVER_VERSIONS contains latest version in all active datacenter branches
+# NOTE: the ubuntu2604 CUDA repository does not carry branch 580, so the aggregate
+# ubuntu26.04 targets cannot build it; use per-version targets (e.g.
+# build-ubuntu26.04-595.71.05) or override DRIVER_VERSIONS for that distribution.
```
tariq1890: I'd rather we remove this. If we want to persist this context in some form. We can consider adding it to the commit description, but even that is not really needed IMO

.github/workflows/image.yaml
```diff
@@ -72,7 +77,11 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v4
         with:
-          image: tonistiigi/binfmt:qemu-v6.2.0
+          # The Ubuntu 26.04 arm64 userland needs a newer emulator than the qemu-v6.2.0
+          # used for every other distribution: v6.2.0 hangs on it, and on the
+          # resolute-20260707 userland v8.1.5 fails with ENOSYS and v9.2.2 with EINVAL
+          # (tar file extraction), so it is pinned to qemu-v10.1.3.
```
tariq1890, line 82: This is a good example of where a multi-line comment is justified 👍

.gitlab-ci.yml
rajathagasthya: All the existing targets have it, but you're right, it's not necessary.

tariq1890 (approved):
Thanks a lot @rajathagasthya !

PR #860, opened 2026-07-08
[precompiled]: rebuild/overwrite the image even if the tag already exists

### LINKED ISSUE #861: [precompiled]: rebuild/overwrite the image even if the tag already exists
Add FORCE_PUSH override to overwrite already-published precompiled images

Issue: When a CVE is found during a release and the precompiled image is already published, the pipelines skip re-pushing because the image tag already exists. If no new kernel has been released, there is no way to rebuild and overwrite the fixed image

Solution: Write the existing FORCE_PUSH into both scheduled pipelines(github and gitlab) so it can be toggled via a CI/CD variable.
Setting FORCE_PUSH=true forces a rebuild, e2e test, and overwrite of the existing tag.
Unset keeps the normal skip.

shivakunv:
Add FORCE_PUSH override to overwrite already-published precompiled images

Issue: When a CVE is found during a release and the precompiled image is already published, the pipelines skip re-pushing because the image tag already exists. If no new kernel has been released, there is no way to rebuild and overwrite the fixed image 

Solution: Write the existing FORCE_PUSH into both scheduled pipelines(github and gitlab) so it can be toggled via a CI/CD variable. 
Setting FORCE_PUSH=true forces a rebuild, e2e test, and overwrite of the existing tag.
Unset keeps the normal skip.

### DIFF
--- a/.github/workflows/precompiled.yaml
+++ b/.github/workflows/precompiled.yaml
@@ -200,6 +200,9 @@ jobs:
           DIST: ${{ matrix.dist }}
           GH_TOKEN:  ${{ secrets.GITHUB_TOKEN }}
           LTS_KERNEL: ${{ matrix.lts_kernel }}
+          # Set FORCE_PUSH=true (Settings > Secrets and variables > Actions > Variables)
+          # to rebuild/e2e/overwrite already-published precompiled images
+          FORCE_PUSH: ${{ vars.FORCE_PUSH }}
         run: |
           kernel_flavors_json='${{ needs.set-driver-version-matrix.outputs.kernel_flavors }}'
           KERNEL_FLAVORS=($(echo "$kernel_flavors_json" | jq -r '.[]'))
--- a/tests/scripts/findkernelversion.sh
+++ b/tests/scripts/findkernelversion.sh
@@ -31,13 +31,18 @@ if [ -n "$artifact" ]; then
     fi
 fi
 
-# calculate driver tag
-status_nvcr=0
-status_ghcr=0
-regctl tag ls  nvcr.io/nvidia/driver | grep "^${DRIVER_BRANCH}-${KERNEL_VERSION}-${DIST}$" || status_nvcr=$?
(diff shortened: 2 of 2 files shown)


shivakunv:
passed pipeline:https://github.com/NVIDIA/gpu-driver-container/actions/runs/28933848126

rajathagasthya:
Let's not do this. We never want to overwrite tags. If there are CVE fixes, they will be carried in the next precompiled driver image. This is how it works for non-precompiled images as well.

shivakunv:
> Let's not do this. We never want to overwrite tags. If there are CVE fixes, they will be carried in the next precompiled driver image. This is how it works for non-precompiled images as well.



_Why force write is necessary_

1. Non-precompiled images: these go through testing on every release, we fix the critical bug and then release. So, a CVE fix naturally lands in the next release.

2. Precompiled images: a precompiled release happens only with a new kernel release, and precompiled testing is done only for a major release:

       a) A new kernel version is detected automatically (proceed to (b) and (c) only if the tag does not exist in both ghcr and nvcr).
       b) github pipeline runs build -> e2e -> publish to ghcr.
       c) gitlab pipeline pushes it to nvstaging (only if the tag does not exist).
       d) nspect manual MR creation and publishing of images.

3. Problem:  
       because the tag already exists, this image tag and this specific kernel version of the GPU driver container will never be published in github pipeline. The buggy image (CVE or scan failure) will stay in nvstaging permanently, and we will miss that kernel release.

5. Waiting does not reliably fix it: 
    if we wait for the next kernel (normally 10–15 days) and it again has a scan failure, we have to wait again and we may end up skipping 3–4 kernel versions this way.

6. Scope is limited to staging: 
    this force write only writes to nvstaging. It is never pushed to the public nvcr registry and does not update the actual release on nvcr.

7. The gitlab pipeline already has this CI/CD variable to overwrite the image.

8. With this change, we no longer miss a new kernel release.

9. `FORCE_PUSH` runs the normal github pipeline job (`build -> e2e -> publish`), the only difference is that it overwrites the buggy image already sitting in nvstaging (in this case ubuntu26.04)


_If you agree, I will add an vuln exception for this CVE and proceed with releasing the precompiled driver container. We can address the CVE in the next release. Please let me know your suggestion._

PR #873, opened 2026-07-20
Fix precompiled image publish silently dropping non-native-arch manifests

### LINKED ISSUE #877: Skopeo multi-arch publish issue
issue: oci-archive: silently drops arm64 manifests , while oci: (directory) with --all preserves the full amd64+arm64 OCI index.

The oci-archive:/oci: transports only support some shapes of indices / manifests — as a guess, a multi-platform image directly stored in the OCI format as a top-level index is not supported; it needs to be an index which contains a nested multi-platform image index.

Solution :
Replace skopeo copy oci-archive:file.tar  with tar -xf + skopeo copy --all oci:dir to correctly publish multi-arch images.

shivakunv:
The publish step for precompiled driver images used `skopeo copy oci-archive:file.tar docker://...` without `--all`. For a multi-platform (amd64+arm64) OCI archive, skopeo doesn't error on that, it silently collapses to a single current-arch manifest, so the pushed image only had one architecture even though the build produced both.

Fix: add `--all` to the existing `oci-archive:` copy. I verified locally against the exact skopeo version CI installs (1.13.3+ds1-2ubuntu0.24.04.3, pulled from a real scheduled run's logs) and against a real push to `ghcr.io` that `--all` alone is enough, no need to extract the archive to an `oci:` directory first like the original fix did.

Also kept and adjusted the post-publish verification, since a `skopeo copy` that exits 0 isn't proof the full index was published (that's exactly how this bug shipped unnoticed in the first place). It compares the local artifact's manifest against what's actually live in the registry, and now handles both shapes correctly: an index (the multi-arch legs) compares per-platform digests, a plain manifest (ubuntu22.04, ubuntu26.04, and ubuntu24.04 azure-fde, which build amd64-only per `multi-arch.mk`) compares config/layer digests instead of silently passing on `[] == []`.

One correction on the linked skopeo issue: that issue is a different, still-open failure (a hard crash on a flat index with no attestation manifests), not the silent-drop behavior described above. Related context, but not the exact bug we're hitting here, that's just the missing `--all` flag.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -23,6 +23,14 @@ on:
       - release-*
 
 jobs:
+  test-verify-manifest-match:
+    runs-on: linux-amd64-cpu4
+    steps:
+      - uses: actions/checkout@v7
+        name: Check out code
+      - name: Run verify-manifest-match tests
+        run: ./.github/scripts/test-verify-manifest-match.sh
+
   image:
     runs-on: linux-amd64-cpu4
     strategy:
--- a/.github/workflows/precompiled.yaml
+++ b/.github/workflows/precompiled.yaml
@@ -492,9 +492,29 @@ jobs:
         if: steps.set_image_vars.outputs.run_publish == 'true'
         run: |
           image_path="./driver-images-${{ matrix.driver_branch }}-${{ env.KERNEL_VERSION }}.tar"
+          IMAGE="${PRIVATE_REGISTRY}/nvidia/driver:${{ matrix.driver_branch }}-${{ env.KERNEL_VERSION }}"
           echo "uploading  $image_path"
           if [[ "${{ github.ref == 'refs/heads/main' }}" == "true" ]]; then
-            skopeo copy --authfile "${REGISTRY_AUTH_FILE}" "oci-archive:${image_path}" docker://${PRIVATE_REGISTRY}/nvidia/driver:${{ matrix.driver_branch }}-${{ env.KERNEL_VERSION }}
+            mkdir -p /tmp/oci-image
+            tar -xf "${image_path}" -C /tmp/oci-image
+            skopeo copy --all --authfile "${REGISTRY_AUTH_FILE}" "oci:/tmp/oci-image" docker://${IMAGE}
+
+            # Verify published manifest matches the built OCI artifact by comparing
+            # platform digests — ensures skopeo copy pushed all manifests correctly
+            LOCAL=$(skopeo inspect --raw "oci:/tmp/oci-image" | \
+              jq -c '[.manifests[]? | {arch: .platform.architecture, digest: .digest}] | sort_by(.arch)')
+            REMOTE=$(skopeo inspect --raw "docker://${IMAGE}" | \
+              jq -c '[.manifests[]? | {arch: .platform.architecture, digest: .digest}] | sort_by(.arch)')
+            if [[ "$LOCAL" == "$REMOTE" ]]; then
+              echo "OK: published manifest matches built artifact"
+            else
+              echo "ERROR: manifest mismatch between artifact and published image"
+              echo "Local:  $LOCAL"
+              echo "Remote: $REMOTE"
+              exit 1
+            fi
+
+            rm -rf /tmp/oci-image
           else
             echo "Skipping image push for non-main branch ${{ github.ref }}"
           fi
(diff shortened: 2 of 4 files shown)


rajathagasthya:
@shivakunv Can you update the PR title and description to better reflect what you're doing in this PR?

shivakunv:
> @shivakunv Can you update the PR title and description to better reflect what you're doing in this PR?

updated

abrarshivani:
Verification:

1. `skopeo copy oci-archive:file.tar docker://...` without `--all` doesn't error on a multi-platform (amd64+arm64) archive — it silently collapses to a single manifest, matching whichever architecture the copy happened to run on. The registry only ever received one architecture, with nothing to flag it.

2. Reproduced this locally with the exact skopeo version CI installs (`1.13.3+ds1-2ubuntu0.24.04.3`, from a real scheduled run's logs), in a matching Ubuntu 24.04 container, against a real multi-arch OCI archive built the same way this workflow builds one.

3. Confirmed the fix: adding `--all` to the existing `oci-archive:` copy publishes the full manifest list for both architectures. Extracting to an `oci:` directory first, as the original approach did, turned out not to be necessary.

4. Pushed for real to `ghcr.io` (the actual destination registry), to a scratch tag under a personal namespace, and ran the verification logic against it — local and remote matched exactly for both `amd64` and `arm64`.

5. Not yet verified: the real `Publish image` step is gated to `github.ref == 'refs/heads/main'`, so nothing here has run inside an actual GitHub Actions job. First real execution is the next scheduled run on `main` after merge.

sylvesterkaczmarek (commented):
The index comparison keeps only `architecture` and `digest`, so platform metadata such as `os` and `variant` is ignored. Those fields live on the index descriptor rather than the child manifest digest, so a published index can carry the right digest under the wrong platform descriptor and still pass this check. Compare the full relevant platform tuple plus digest and add an ARM variant regression.

abrarshivani:
@sylvesterkaczmarek Thanks for catching this. Updated the comparison to include the full platform tuple (`os`, `architecture`, `variant`) along with the digest, and added a regression test that verifies crossed variant/digest mappings are rejected.

.github/workflows/image.yaml
```diff
@@ -23,18 +23,24 @@ on:
       - release-*
 
 jobs:
+  test-verify-manifest-match:
+    runs-on: linux-amd64-cpu4
+    steps:
+      - uses: actions/checkout@v7
+        name: Check out code
+      - name: Run verify-manifest-match tests
```
rajathagasthya, line 31: I'm wondering if running this test is really necessary. Not a strong opinion, so if you want to keep it that's fine.

rajathagasthya (approved):
Thanks for identifying the root cause @abrarshivani!

.github/workflows/image.yaml
abrarshivani: I think it is good to have this and plan to keep it for now.

sylvesterkaczmarek (commented):
Re-reviewed the platform-identity finding. Index summaries now compare OS, architecture, variant and digest, and the added ARM-variant regression covers the crossed-variant case I raised. My previous blocker is resolved.

.github/workflows/precompiled.yaml
```diff
@@ -492,16 +492,21 @@ jobs:
...
           image_path="./driver-images-${{ matrix.driver_branch }}-${{ env.KERNEL_VERSION }}.tar"
+          IMAGE="${PRIVATE_REGISTRY}/nvidia/driver:${{ matrix.driver_branch }}-${{ env.KERNEL_VERSION }}"
           echo "uploading  $image_path"
           if [[ "${{ github.ref == 'refs/heads/main' }}" == "true" ]]; then
-            skopeo copy --authfile "${REGISTRY_AUTH_FILE}" "oci-archive:${image_path}" docker://${PRIVATE_REGISTRY}/nvidia/driver:${{ matrix.driver_branch }}-${{ env.KERNEL_VERSION }}
+            skopeo copy --all --authfile "${REGISTRY_AUTH_FILE}" "oci-archive:${image_path}" "docker://${IMAGE}"
+
+            skopeo inspect --raw "oci-archive:${image_path}" > ./local-manifest.json
+            skopeo inspect --raw --authfile "${REGISTRY_AUTH_FILE}" "docker://${IMAGE}" > ./remote-manifest.json
+            .github/scripts/verify-manifest-match.sh ./local-manifest.json ./remote-manifest.json
```
tariq1890, line 502: should we clean these files up after the test is run or is it unnecessary?

tariq1890 (approved):

.github/workflows/precompiled.yaml
abrarshivani: I don't think cleanup is necessary here. These are temporary files in the job workspace and aren't used after the verification. The workspace is also cleaned on the next checkout, so I would leave them as is.

sylvesterkaczmarek:
Thanks for addressing the platform-identity edge case and for merging this. I appreciated the full OS/architecture/variant regression coverage.

PR #893, opened 2026-07-31
add support for overriding fields in fabricmanager.cfg

tariq1890:
Allow gpu-driver-container users to override fabric manager config before starting the fabricmanager daemon on their NVSwitch GPU systems. With this change, the gpu-driver-container will the parse environment variables with the `NVFM_CONFIG_` prefix and modify the `/usr/share/nvidia/nvswitch/fabricmanager.cfg` file with the desired fabric manager config before starting up fabricmanager.

### DIFF
--- a/rhel10/nvidia-driver
+++ b/rhel10/nvidia-driver
@@ -708,6 +708,38 @@ _start_vgpu_topology_daemon() {
     nvidia-topologyd
 }
 
+_configure_fabric_manager() {
+    local fm_config_file=/usr/share/nvidia/nvswitch/fabricmanager.cfg
+
+    local nvfm_env_vars
+    nvfm_env_vars=$(env | grep '^NVFM_CONFIG_' || true)
+    [ -z "${nvfm_env_vars}" ] && return 0
+
+    if [ ! -f "${fm_config_file}" ]; then
+        echo "WARNING: ${fm_config_file} not found; cannot configure Fabric Manager"
+        return 0
+    fi
+
+    while IFS='=' read -r name value; do
+        local field="${name#NVFM_CONFIG_}"
+        if [ -z "${field}" ]; then
+            echo "WARNING: ignoring environment variable '${name}=' with an empty config field name"
+            continue
+        fi
+        echo "Setting Fabric Manager config ${field}=${value} in ${fm_config_file}"
+        if grep -q "^${field}=" "${fm_config_file}"; then
+            # Escape characters that are special in the replacement half of a sed
+            # s||| command (backslash, ampersand and the '|' delimiter) so that a
+            # value cannot corrupt the substitution or inject extra sed commands.
+            local sed_escaped_value
+            sed_escaped_value=$(printf '%s' "${value}" | sed -e 's/[\\&|]/\\&/g')
+            sed -i "s|^${field}=.*|${field}=${sed_escaped_value}|" "${fm_config_file}"
+        else
+            printf '\n%s=%s\n' "${field}" "${value}" >> "${fm_config_file}"
+        fi
+    done <<< "${nvfm_env_vars}"
+}
+
 _start_daemons() {
     echo "Starting NVIDIA persistence daemon..."
     nvidia-persistenced --persistence-mode
@@ -730,6 +762,7 @@ _start_daemons() {
 
     if _assert_nvlink5_system; then
         _ensure_nvlink5_prerequisites || return 1
+        _configure_fabric_manager
         echo "Starting NVIDIA fabric manager daemon for NVLink5+..."
 
         fm_config_file=/usr/share/nvidia/nvswitch/fabricmanager.cfg
@@ -744,6 +777,7 @@ _start_daemons() {
 
     # If not a NVLink5+ switch, check for the presence of NVLink4 (or below) switches
     elif _assert_nvswitch_system; then
+        _configure_fabric_manager
         echo "Starting NVIDIA fabric manager daemon..."
         nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
     fi
--- a/rhel8/nvidia-driver
+++ b/rhel8/nvidia-driver
@@ -685,6 +685,38 @@ _start_vgpu_topology_daemon() {
     nvidia-topologyd
 }
 
+_configure_fabric_manager() {
+    local fm_config_file=/usr/share/nvidia/nvswitch/fabricmanager.cfg
+
+    local nvfm_env_vars
+    nvfm_env_vars=$(env | grep '^NVFM_CONFIG_' || true)
+    [ -z "${nvfm_env_vars}" ] && return 0
+
+    if [ ! -f "${fm_config_file}" ]; then
+        echo "WARNING: ${fm_config_file} not found; cannot configure Fabric Manager"
+        return 0
+    fi
+
+    while IFS='=' read -r name value; do
+        local field="${name#NVFM_CONFIG_}"
+        if [ -z "${field}" ]; then
+            echo "WARNING: ignoring environment variable '${name}=' with an empty config field name"
+            continue
+        fi
+        echo "Setting Fabric Manager config ${field}=${value} in ${fm_config_file}"
+        if grep -q "^${field}=" "${fm_config_file}"; then
+            # Escape characters that are special in the replacement half of a sed
+            # s||| command (backslash, ampersand and the '|' delimiter) so that a
+            # value cannot corrupt the substitution or inject extra sed commands.
+            local sed_escaped_value
+            sed_escaped_value=$(printf '%s' "${value}" | sed -e 's/[\\&|]/\\&/g')
+            sed -i "s|^${field}=.*|${field}=${sed_escaped_value}|" "${fm_config_file}"
+        else
+            printf '\n%s=%s\n' "${field}" "${value}" >> "${fm_config_file}"
+        fi
+    done <<< "${nvfm_env_vars}"
+}
+
 _start_daemons() {
     echo "Starting NVIDIA persistence daemon..."
     nvidia-persistenced --persistence-mode
@@ -707,6 +739,7 @@ _start_daemons() {
 
     if _assert_nvlink5_system; then
         _ensure_nvlink5_prerequisites || return 1
+        _configure_fabric_manager
         echo "Starting NVIDIA fabric manager daemon for NVLink5+..."
 
         fm_config_file=/usr/share/nvidia/nvswitch/fabricmanager.cfg
@@ -721,6 +754,7 @@ _start_daemons() {
 
     # If not a NVLink5+ switch, check for the presence of NVLink4 (or below) switches
     elif _assert_nvswitch_system; then
+        _configure_fabric_manager
         echo "Starting NVIDIA fabric manager daemon..."
         nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg
     fi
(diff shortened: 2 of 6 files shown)


varunrsekar:
@tariq1890 In the interest of just tweaking fabric-mode, this PR looks correct to me. 

But what about other configuration tweaks like the resiliency mode and the FM service connection method? Are they going to be handled in separate PRs?

tariq1890:
Thanks for taking a look @varunrsekar. I have extended the implementation to make it more generic. Now, users can override the fabric manager config field of their choice by providing the environment variables with this format - `NVFM_CONFIG_<FABRIC_MANAGER_CONFIG_FIELD>=<FABRIC_MANAGER_CONFIG_VALUE>`

cdesiniotis (approved):

tariq1890:
Sharing a snippet of the driver container logs after this change

```
....
....
Parsing kernel module parameters...
Configuring nvidia module parameters in /etc/modprobe.d/nvidia.conf
Installing NVIDIA driver kernel modules...

WARNING: The nvidia-drm module will not be installed. As a result, DRM-KMS will not function with this installation of the NVIDIA driver.


ERROR: Unable to open 'kernel-open/dkms.conf' for copying (No such file or directory)


Welcome to the NVIDIA Software Installer for Unix/Linux

Detected 128 CPUs online; setting concurrency level to 32.
Unable to locate any tools for listing initramfs contents.
Unable to scan initramfs: no tool found
Installing NVIDIA driver version 595.71.05.
Performing CC sanity check with CC="/usr/bin/cc".
Performing CC check.
Kernel source path: '/lib/modules/5.15.0-186-generic/build'

Kernel output path: '/lib/modules/5.15.0-186-generic/build'

Performing Compiler check.
Performing Dom0 check.
Performing Xen check.
Performing PREEMPT_RT check.
Performing vgpu_kvm check.
Cleaning kernel module build directory.
Building kernel modules:

  [##############################] 100%
Kernel module compilation complete.
Installing 'NVIDIA Accelerated Graphics Driver for Linux-x86_64' (595.71.05):: Installing

  [                              ]   0%
Unable to determine whether NVIDIA kernel modules are present in the initramfs. Existing NVIDIA kernel modules in the initramfs, if any, may interfere with the newly installed driver.

  [##############################] 100%
Driver file installation is complete.
Running post-install sanity check:: Checking

  [##############################] 100%
Post-install sanity check passed.

Installation of the kernel module for the NVIDIA Accelerated Graphics Driver for Linux-x86_64 (version: 595.71.05) is now complete.

Configuring the following firmware search path in '/sys/module/firmware_class/parameters/path': /run/nvidia/driver/lib/firmware
WARNING: A search path is already configured in /sys/module/firmware_class/parameters/path
         Retaining the current configuration
Loading ipmi and i2c_core kernel modules...
Loading NVIDIA driver kernel modules...
+ modprobe nvidia
+ modprobe nvidia-uvm
+ modprobe nvidia-modeset
+ set +o xtrace -o nounset
Setting Fabric Manager config LOG_LEVEL=2 in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Setting Fabric Manager config LOG_MAX_ROTATE_COUNT=5 in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Setting Fabric Manager config LOG_FILE_MAX_SIZE=2048 in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Setting Fabric Manager config FABRIC_MODE=1 in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Starting NVIDIA persistence daemon...
Starting NVIDIA fabric manager daemon...
Mounting NVIDIA driver rootfs...
Storing driver configuration digest...
Driver configuration digest stored at /run/nvidia/nvidia-driver.state
Done, now waiting for signal
```

varunrsekar (approved):
Change looks good to me!

rhel10/nvidia-driver
```diff
@@ -708,6 +709,34 @@ _start_vgpu_topology_daemon() {
...
+        echo "Setting Fabric Manager config ${field}=${value} in ${fm_config_file}"
+        if grep -q "^${field}=" "${fm_config_file}"; then
+            # Escape characters that are special in the replacement half of a sed
+            # s||| command (backslash, ampersand and the '|' delimiter) so that a
+            # value cannot corrupt the substitution or inject extra sed commands.
+            local sed_escaped_value
+            sed_escaped_value=$(printf '%s' "${value}" | sed -e 's/[\\&|]/\\&/g')
+            sed -i "s|^${field}=.*|${field}=${sed_escaped_value}|" "${fm_config_file}"
+        else
+            printf '\n%s=%s\n' "${field}" "${value}" >> "${fm_config_file}"
```
shivamerla, line 738: nit: if the env is set as just `NVFM_CONFIG_` this will end up appending with empty key, we could add a check to safe guard this and throw an error at the beginning of this method.

shivamerla, line 738: The default FM config file seems to include all supported keys, even with empty values. We could probably fail fast if an unknown key is configured rather than silently appending it. Just worth confirming with the FM team that there aren't any supported keys that may be intentionally omitted from the default config.

tariq1890: The thing is that could change in the future. I am not sure if it's worth maintaining a static list of all the supported/valid config fields.

If the user provides these config values, it's their responsibility to provide the correct env vars. In the future, we could expose knobs in the CR and those CR fields could be converted into these `NVFM_CONFIG_` prefixed envars. But that can be done only once we have fully thought through the API and UX changes we want to make.

tariq1890: Fair point. What if we just ignored that env var instead of erroring out?

shivamerla: yes, that sounds good as well.

shivamerla: I mean the default config file installed via the fabric manager package already have all supported fields. We don't have to statically maintain it. If say a new field is supported with latest FM version x.y then we silently add those fields in older versions as well, not sure what is the behavior there, whether it will ignore those or fail during startup as unknown field?

rhel10/nvidia-driver
```diff
@@ -431,6 +431,7 @@ _load_driver() {
         set +o xtrace -o nounset
     fi
 
+    _configure_fabric_manager || exit 1
```
rahulait: Should we always run it? I am thinking of an edge case where the user accidentally sets `NVFM_CONFIG_` env var on a system with no nvlink and we'll keep erroring with message that the config file doesn't exist, we cannot configure fabric manager.

tariq1890: We always install the FabricManager at build time. It will always exist regardless of the system having an NVSwitch. The error block was just added for "completeness", but we don't ever expect that code path to get traversed during a driver container run time

rhel10/nvidia-driver
tariq1890: This has been done. Please check now

tariq1890: > If say a new field is supported with latest FM version x.y then we silently add those fields in older versions as well, not sure what is the behavior there, whether it will ignore those or fail during startup as unknown field?

This implementation assumes that the user is intentional about the values they pass in and that they are aware of the driver version that is being used while doing so.

shivamerla: Can you verify what happens if an unknown config parameter is provided? I'm curious whether FM ignores it gracefully or treats it as an error. One case to consider is when a user upgrades the driver version in the NVIDIADriver CR but leaves the existing environment variables unchanged. We should confirm the behavior and document it. If FM silently ignores unknown parameters, it could lead to configuration drift that's not obvious, leaving users to debug why their expected configuration wasn't applied.

rhel10/nvidia-driver
tariq1890: I guess I could have it emit a warning instead of erroring out if the config file doesn't; given that we don't realistically expect the `|| exit` to ever execute.

tariq1890: Updated

rhel10/nvidia-driver
tariq1890: yes, will do

rhel10/nvidia-driver
rahulait: Thx, we do check whether the system has nvlink switches inside start_daemons with _assert_nvlink5_system and _assert_nvswitch_system calls. So was wondering if we should use them here as well. Making the check to not fail and just log warning is also fine.

rhel10/nvidia-driver
tariq1890: Looks like invalid configuration leads to a startup failure of the FM daemon

```
.....
Loading NVIDIA driver kernel modules...
+ modprobe nvidia
+ modprobe nvidia-uvm
+ modprobe nvidia-modeset
+ set +o xtrace -o nounset
Setting Fabric Manager config FM_CMD_UNIX_SOCKET_PATH=/run/nvidia-fabricmanager/socket in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Setting Fabric Manager config RANDOM_FIELD=randomvalue in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Setting Fabric Manager config FABRIC_MODE=1 in /usr/share/nvidia/nvswitch/fabricmanager.cfg
Starting NVIDIA persistence daemon...
Starting NVIDIA fabric manager daemon...
unsupported config item RANDOM_FIELD is specified in fabric manager config file
....
```

rhel10/nvidia-driver
tariq1890: > So was wondering if we should use them here as well.

I did consider that initially, but I felt it may be too redundant (and potentially wasteful?) to execute those two methods twice.

rhel10/nvidia-driver
shivamerla: ok, we can depend on the FM service to error out in that case then. But looks like you found that the driver-container itself doesn't error out in that case. We need to address that (even as a follow up but critical).

rhel10/nvidia-driver
shivamerla: @tariq1890 why can't we invoke this method from `_start_daemons` itself when we decide to launch FM daemon? That way this only gets executed on environments where this is relevant and required.

rhel10/nvidia-driver
```diff
@@ -708,6 +709,38 @@ _start_vgpu_topology_daemon() {
...
+        echo "WARNING: ${fm_config_file} not found; cannot configure Fabric Manager"
+        return 0
+    fi
+
+    while IFS='=' read -r name value; do
+        local field="${name#NVFM_CONFIG_}"
+        if [ -z "${field}" ]; then
+            echo "WARNING: ignoring environment variable '${name}=' with an empty config field name"
+            continue
+        fi
```
rahulait, line 728: nit: should we add a check to prevent regex expansion for key? Just to ignore something like `NVFM_CONFIG_.*`. Seems like an overkill but just wondering:
```
        if [[ ! "${field}" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]]; then
            echo "WARNING: ignoring '${name}': invalid Fabric Manager config field name"
            continue
        fi
```

rhel10/nvidia-driver
tariq1890: I think I can make that change

rhel10/nvidia-driver
tariq1890: Probably not worth the effort since the FM startup script will immediately fail if it sees an unrecognised config

rhel10/nvidia-driver
tariq1890: @shivamerla This is done. Please check now

shivamerla:
Looks good to me with the changes addressed.

shivamerla (approved):

PR #898, opened 2026-08-03
install the FabricManager SDK packages

tariq1890:
With this change, the driver container images will now download and save the fabricmanager-dev/devel packages during image build. This is needed by the DRA driver that performs dynamic fabric partitioning.

NOTE: At the time of filing this PR, the oldest supported driver branch is R580, so this PR limits the scope of the package install scripts to drivers starting from R580 and newer

### DIFF
--- a/rhel10/install.sh
+++ b/rhel10/install.sh
@@ -152,23 +152,20 @@ nvidia_installer () {
 }
 
 fabricmanager_install() {
-  local fabricmanager_package_name=nvidia-fabricmanager
-  dnf install -y ${fabricmanager_package_name}-${DRIVER_VERSION}
-  dnf versionlock add ${fabricmanager_package_name}
+  dnf install -y nvidia-fabricmanager-${DRIVER_VERSION} nvidia-fabric-manager-devel-${DRIVER_VERSION}
+  dnf versionlock add nvidia-fabricmanager nvidia-fabric-manager-devel
 }
 
 nscq_install() {
-  local nscq_package_name=libnvidia-nscq
-  dnf install -y ${nscq_package_name}-${DRIVER_VERSION}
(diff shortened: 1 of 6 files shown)


cdesiniotis (approved):

rhel10/install.sh
```diff
@@ -152,23 +152,20 @@ nvidia_installer () {
 }
 
 fabricmanager_install() {
-  local fabricmanager_package_name=nvidia-fabricmanager
-  dnf install -y ${fabricmanager_package_name}-${DRIVER_VERSION}
-  dnf versionlock add ${fabricmanager_package_name}
+  dnf install -y nvidia-fabricmanager-${DRIVER_VERSION} nvidia-fabric-manager-devel-${DRIVER_VERSION}
```
rahulait, line 155: just for my own knowledge: any specific reason we plan to install *-devel package as well. Just wondering why we didn't needed it before and need it now.

rahulait (approved):

rhel10/install.sh
rahulait: > With this change, the driver container images will not download and save the fabricmanager-dev/devel packages during image build. 

I think you meant it will now download.

rajathagasthya (approved):

rhel10/install.sh
tariq1890: We are doing it to enable the DRA driver to communicate with Fabric Manager endpoint. For this to work, the FM SDK package versions have to match the driver version. Given this requirement, we install the FM devel packages in the driver container. The DRA driver once deployed will leverage these installed packages to perform dynamic memory fabric partitioning.

tariq1890: > I think you meant it will now download.

Yes, thanks for pointing that out; I will update the description.

PR #908, opened 2026-08-05
fail fast when FabricManager fails to come up

tariq1890:
Today, the driver containers go into `Running` mode even if Fabric Manager doesn't come up successfully. Given that Fabric Manager is a must-have for GPUs to be usable in NVSwitch-based nodes, the driver container should fail fast instead. This would make it easier for users to debug any driver container issues relating to Fabric Manager as they get immediate feedback


rajathagasthya (approved):

rahulait (approved):
LGTM 🚀

shivamerla:
Looks good to me as well! This needs a good amount of testing to make sure there are no spurious failures with some of these daemons.

PR #909, opened 2026-08-06
[B200+][NVLink5+] reset GPUs after FABRIC_MODE=1 is set

tariq1890:
NVLinked systems that use NVLink generation 5 or greater require the GPUs to be reset after Fabric Manager is started up with `FABRIC_MODE=1` set in its config.

### DIFF
--- a/rhel10/nvidia-driver
+++ b/rhel10/nvidia-driver
@@ -774,6 +774,12 @@ _start_daemons() {
                                                --fm-pid-file $fm_pid_file \
                                                --nvlsm-config-file $nvlsm_config_file \
                                                --nvlsm-pid-file $nvlsm_pid_file || return 1
+        if [ ! -z "${NVFM_CONFIG_FABRIC_MODE:-}" ] && [ "${NVFM_CONFIG_FABRIC_MODE}" == "1" ]; then
+            # We reset the GPU devices after Fabric Manager startup to clear stale FM
+            # partition data after setting FABRIC_MODE=1
+            echo "Resetting GPUs to clear stale fabric partition states..."
+            nvidia-smi -r
+        fi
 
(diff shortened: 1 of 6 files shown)


rhel10/nvidia-driver
```diff
@@ -774,6 +774,10 @@ _start_daemons() {
                                                --fm-pid-file $fm_pid_file \
                                                --nvlsm-config-file $nvlsm_config_file \
                                                --nvlsm-pid-file $nvlsm_pid_file
+        # We reset the GPU devices after Fabric Manager startup to clear stale FM
+        # partition data that may have carried over previous driver container runs
+        echo "Resetting GPUs to clear stale Fabric Partition states..."
+        nvidia-smi -r
```
varunrsekar: Instead of a blanket reset, I wonder if we need to check if the fabric state is `In Progress` first on all GPUs before attempting to reset:
```
nvidia-smi --query-gpu "fabric.state" --format "csv,noheader"
```

tariq1890: Isn't it possible that the `Completed` state could be stale as well?

varunrsekar: My statement was incorrect here. When switch from `FABRIC_MODE` 0 to 1, all the GPUs will go to `In Progress` fabric state until there's a partition activation that uses those GPUs. So its not valid to reset the GPU by checking for the fabric state.

Your current implementation of doing the reset if we're trying to start in `FABRIC_MODE=1` makes sense.

varunrsekar (approved):

rhel10/nvidia-driver
tariq1890: Thanks for confirming! I will resolve this comment.

shivamerla (approved):

PR #913, opened 2026-08-10
add new status file to signal daemons startup

tariq1890:
This PR introduces a new status file called `/run/nvidia/validations/.driver-daemons-status`. This file can be leveraged by components downstream who need to ensure that the driver daemons like persistenced, fabricmanager, vgpu-topologyd have come up before performing any other operations

### DIFF
--- a/rhel10/nvidia-driver
+++ b/rhel10/nvidia-driver
@@ -95,6 +95,11 @@ _resolve_kernel_version() {
     return 0
 }
 
+_set_daemons_status() {
+    mkdir -p /run/nvidia/validations
+    echo "${1}" > /run/nvidia/validations/.driver-daemons-status
+}
+
 # Install the kernel modules header/builtin/order files and generate the kernel version string.
 _install_prerequisites() (
     local tmp_dir=$(mktemp -d)
@@ -787,6 +792,7 @@ _start_daemons() {
         echo "Starting NVIDIA fabric manager daemon..."
         nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg || return 1
     fi
+    _set_daemons_status "Ready"
 }
 
 _store_driver_digest() {
@@ -877,6 +883,7 @@ _load() {
 }
 
 init() {
+    _set_daemons_status "NotReady"
     _prepare_exclusive
 
     if _should_skip_kernel_module_reload; then
--- a/ubuntu24.04/nvidia-driver
+++ b/ubuntu24.04/nvidia-driver
@@ -97,6 +97,11 @@ _resolve_kernel_version() {
     return 0
 }
 
+_set_daemons_status() {
+    mkdir -p /run/nvidia/validations
+    echo "${1}" > /run/nvidia/validations/.driver-daemons-status
+}
+
 # Install the kernel modules header/builtin/order files and generate the kernel version string.
 _install_prerequisites() (
     local tmp_dir=$(mktemp -d)
@@ -303,7 +308,6 @@ _load_driver() {
 
     echo "Loading ipmi and i2c_core kernel modules..."
     modprobe -a i2c_core ipmi_msghandler ipmi_devintf
-
     echo "Loading NVIDIA driver kernel modules..."
     set -o xtrace +o nounset
     modprobe nvidia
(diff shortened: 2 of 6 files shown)


tariq1890:
Related gpu-operator PR - https://github.com/NVIDIA/gpu-operator/pull/2727/changes

ubuntu24.04/nvidia-driver
```diff
@@ -304,6 +308,8 @@ _load_driver() {
     echo "Loading ipmi and i2c_core kernel modules..."
     modprobe -a i2c_core ipmi_msghandler ipmi_devintf
 
+    _create_daemons_status_file
```
shivamerla: why can't we just invoke this at the beginning of `init()` method itself than at multiple places?

ubuntu26.04/nvidia-driver
rahulait, line 885: It should create that daemons file here as well no? Currently, its creating that file when `_should_skip_kernel_module_reload` returns true.

ubuntu24.04/nvidia-driver
tariq1890: We want to create this before we call the `_unmount_rootfs` method. The reason we need to do this in multiple places is because we run `_start_daemons` both during fast-path (kernel build and load skipped) and the normal path.

ubuntu26.04/nvidia-driver
tariq1890: For the normal path, I am calling `_create_daemons_status_file` from within `_load_driver`

rahulait: I see it doing in `_install_prerequisites`

rhel10/nvidia-driver
```diff
@@ -95,6 +95,10 @@ _resolve_kernel_version() {
     return 0
 }
 
+_create_daemons_status_file() {
+    > "${RUN_DIR}/driver/.driver-daemons-status"
```
shivamerla: Rather than creating an empty file, can we initialize it with `NotReady` as the starting state?

tariq1890: Sure, we can. I don't have a strong opinion here

ubuntu24.04/nvidia-driver
shivamerla: Got it, so this currently depends on the `/run/nvidia/driver` bind mount being available before the status file is created. Do we actually need to expose this file on the host, or could we keep it local to the container? Keeping it container-local would remove the bind-mount dependency and make the flow a bit cleaner. Anyway after the bind-mount entire container `/` will be available on the host.

tariq1890: > Do we actually need to expose this file on the host,

The startup-probe.sh script that we maintain in the gpu-operator will reference the `/run/nvidia/driver`, but I guess it could always reference that container-local path. I am just not sure if that's good practice.

shivamerla:
@tariq1890 in the PR description add details about why a sentinel file is not enough in this case and how that causes backward compatibility issue.

rhel10/nvidia-driver
rahulait: I feel adding `NotReady` is better than empty file for anyone checking the file and its content. A `NotReady` gives a clear signal that things are not yet ready than an empty file which is implementation dependent.

ubuntu24.04/nvidia-driver
tariq1890: @shivamerla I'm creating the files under `/run/nvidia/validations`. I will update the gpu-operator PR accordingly

rhel10/nvidia-driver
tariq1890: This is done. I'll resolve the comment

ubuntu24.04/nvidia-driver
tariq1890: This change is done. 

I've validated them in an Ubuntu 26.04 environment along with the gpu-operator changes. They are working as expected

rhel10/nvidia-driver
```diff
@@ -787,6 +792,7 @@ _start_daemons() {
         echo "Starting NVIDIA fabric manager daemon..."
         nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg || return 1
     fi
+    echo "Ready" > /run/nvidia/validations/.driver-daemons-status
```
shivamerla: can you add a method for this too, `set_daemon_status "Ready"`

tariq1890: This is done

shivamerla:
These changes look good to me.

shivamerla (approved):

PR #971, opened 2026-08-14
[precompiled]: Add support for nvidia-64k precompiled driver container

### LINKED ISSUE #2751: [precompiled]: Add support for nvidia-64k precompiled driver container
[precompiled]: Add support for nvidia-64k precompiled driver container

shivakunv:
Add support for nvidia-64k precompiled driver container

### DIFF
--- a/.github/precompiled-matrix-config.json
+++ b/.github/precompiled-matrix-config.json
@@ -1,6 +1,6 @@
 {
   "driver_branch": ["580", "595"],
-  "kernel_flavors": ["aws", "azure", "azure-fde", "generic", "nvidia", "oracle"],
+  "kernel_flavors": ["aws", "azure", "azure-fde", "generic", "nvidia", "nvidia-64k", "oracle"],
   "dist": ["ubuntu22.04", "ubuntu24.04", "ubuntu26.04"],
   "lts_kernel": ["5.15", "6.8", "7.0"],
   "platforms": ["amd64", "arm64"],
@@ -14,7 +14,10 @@
     { "lts_kernel": "5.15", "dist": "ubuntu26.04" },
     { "lts_kernel": "6.8", "dist": "ubuntu26.04" },
     { "lts_kernel": "7.0", "dist": "ubuntu22.04" },
-    { "flavor": "azure-fde", "dist": "ubuntu22.04" }
+    { "flavor": "azure-fde", "dist": "ubuntu22.04" },
+    { "flavor": "nvidia-64k", "dist": "ubuntu22.04" },
+    { "flavor": "nvidia-64k", "dist": "ubuntu26.04" },
+    { "flavor": "nvidia-64k", "lts_kernel": "5.15" }
   ],
   "exclude_precompiled_e2e_matrix": [
     { "lts_kernel": "5.15", "dist": "ubuntu24.04" },
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -54,11 +54,20 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         apt-utils git curl && \
     rm -rf /var/lib/apt/lists/*
 
-RUN echo "deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ noble main universe" > /etc/apt/sources.list && \
-    echo "deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ noble-updates main universe" >> /etc/apt/sources.list && \
-    echo "deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ noble-security main universe" >> /etc/apt/sources.list && \
-    echo "deb [arch=amd64] http://us.archive.ubuntu.com/ubuntu noble-updates main restricted" >> /etc/apt/sources.list && \
-    echo "deb [arch=amd64] http://us.archive.ubuntu.com/ubuntu noble-security main restricted" >> /etc/apt/sources.list && \
+RUN ARCH=$(dpkg --print-architecture) && \
+    # kernel version is the same for amd64 and arm64, so base image is always built for amd64 ( non-nvidia-64k)
+    # for nvidia-64k is arm64-only, so its base image must be built for arm64
+    if [ "$ARCH" = "arm64" ]; then \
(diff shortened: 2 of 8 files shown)


base/Dockerfile
```diff
@@ -54,11 +54,20 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-    echo "deb [arch=amd64] http://us.archive.ubuntu.com/ubuntu noble-updates main restricted" >> /etc/apt/sources.list && \
-    echo "deb [arch=amd64] http://us.archive.ubuntu.com/ubuntu noble-security main restricted" >> /etc/apt/sources.list && \
+RUN ARCH=$(dpkg --print-architecture) && \
+    # kernel version is the same for amd64 and arm64, so base image is always built for amd64 ( non-nvidia-64k)
+    # for nvidia-64k is arm64-only, so its base image must be built for arm64
+    if [ "$ARCH" = "arm64" ]; then \
+      echo "deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ noble main universe" > /etc/apt/sources.list && \
+      echo "deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ noble-updates main universe" >> /etc/apt/sources.list && \
+      echo "deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports/ noble-security main universe" >> /etc/apt/sources.list; \
+    else \
```
abrarshivani, line 64: Do we need to override the arm64 apt sources here? The noble arm64 base image already has `/etc/apt/sources.list.d/ubuntu.sources` configured for `ports.ubuntu.com` with `main universe restricted multiverse`.

The new arm64 entries only enable `main universe`, while packages used by this flow such as `nvidia-utils-580-server` and `linux-signatures-nvidia-*` are in `restricted`. I think we should either rely on the existing `ubuntu.sources` or include `restricted` explicitly?

Ref:
https://packages.ubuntu.com/noble-updates/linux-modules-nvidia-595-server-6.8.0-1061-nvidia-64k

base/generate-ci-config
```diff
@@ -26,7 +26,12 @@ DRIVER_VERSION=$(apt-cache show nvidia-utils-${DRIVER_BRANCH}-server |grep Versi
 # Get the latest kernel from linux-signatures-list and linux-images-list
 # As list is already sorted , compare the kernel version and find exact match
 # get the latest kernel version with tail
-SK=$(grep -Fxf <(echo "$SUPPORTED_KERNELS_LINUX_SIGNATURES_LIST") <(echo "$SUPPORTED_KERNELS_LINUX_IMAGE_LIST") | tail -n1)
+# linux-signatures-nvidia is not available for arm64 (e.g. nvidia-64k flavor); fall back to linux-image list only
+if [ -n "$SUPPORTED_KERNELS_LINUX_SIGNATURES_LIST" ]; then
+  SK=$(grep -Fxf <(echo "$SUPPORTED_KERNELS_LINUX_SIGNATURES_LIST") <(echo "$SUPPORTED_KERNELS_LINUX_IMAGE_LIST") | tail -n1)
+else
```
abrarshivani: It looks like `linux-signatures-nvidia-*` is available for `nvidia-64k` on arm64, so this fallback may not be necessary. I think it is good have the existing image/signature intersection and fail explicitly if no matching kernel is found, rather than falling back to the latest `linux-image`.
Ref:
https://packages.ubuntu.com/noble-updates/linux-modules-nvidia-595-server-6.8.0-1061-nvidia-64k

scripts/precompiled.sh
```diff
@@ -76,12 +83,20 @@ function pushImage(){
...
+        echo "nvidia-64k is arm64-only, skipping amd64 pull"
+        return 0
+    fi
     make DRIVER_VERSIONS=${DRIVER_VERSIONS} DRIVER_BRANCH=${DRIVER_BRANCH} pull-${DIST}-${DRIVER_VERSION}
 }
 
 function archiveImage(){
-    # archive the image
+    # nvidia-64k is arm64-only; skip if scanning for amd64
+    if [[ "${KERNEL_FLAVOR}" == "nvidia-64k" ]] && [[ "${PLATFORM}" == "linux/amd64" ]]; then
```
abrarshivani: For `nvidia-64k` on the amd64 scan path, returning successfully (returns 0) here skips the pull, and `archiveImage()` similarly skips creating `archive.tar`. However, `.scan-generic` still continues and runs `pulse-cli -i archive.tar`, so this job will fail on a missing archive. Can we exclude the `nvidia-64k`/amd64 combination from the scan job instead?

.github/precompiled-matrix-config.json
```diff
@@ -1,6 +1,6 @@
 {
   "driver_branch": ["580", "595"],
-  "kernel_flavors": ["aws", "azure", "azure-fde", "generic", "nvidia", "oracle"],
+  "kernel_flavors": ["aws", "azure", "azure-fde", "generic", "nvidia", "nvidia-64k", "oracle"],
```
abrarshivani, line 3: `nvidia-64k` is not included in `image.yaml`, so this path is not covered by the PR CI. Can we add it to the precompiled matrix in `image.yaml` so it is tested on every PR?

abrarshivani, line 3: Can we make sure the `nvidia-64k` arm64 e2e completes successfully on `g5g.xlarge`?

shivakunv: It is not needed, the nvidia-64k version is supported only on arm64.
The precompiled pipeline already performs this job and runs daily.

shivakunv: please clarify your question?
If you mean how to test this new change and ensure the CI pipeline succeeds before merging, you can always retry it using a manual trigger:

Action->Precompiled images -> Run Workflow -> select branch (support-arm-64k)

abrarshivani: I see the other arm64 precompiled images are covered in PR CI through `image.yaml`. Is there a reason not to include `nvidia-64k` there as an arm64-only case? That will also give us a pre-merge signal for this path.

abrarshivani: Yes, that is what I meant. A manual trigger is fine. I just want to confirm that the nvidia-64k arm64 e2e passes before this change is merged.

shivakunv: fair enough. done

PR #986, opened 2026-08-26
Guard unzboot copy by arch in ocp_dtk_entrypoint, amd64 doesn't have the binary

### LINKED ISSUE #984: [Bug]: DTK mode fails on amd64 - shared-dir prep unconditionally copies /usr/bin/unzboot which is only installed on aarch64
**Describe the bug**

In Driver Toolkit (DTK) mode, `ocp_dtk_entrypoint`'s shared-directory preparation unconditionally copies `/usr/bin/unzboot`. `install.sh` only installs `unzboot` on `aarch64` (EPEL, zboot kernels). On amd64 the file doesn't exist, so the `cp` fails and the driver pod crash-loops before any build starts.

**To Reproduce**

Deploy a current-generation `*-rhel10.0` driver image via gpu-operator with `use_ocp_driver_toolkit: true` on amd64. `nvidia-driver-ctr` immediately fails:

```
+ cp -r /tmp/install.sh /usr/local/bin/ocp_dtk_entrypoint /usr/local/bin/nvidia-driver \
       /usr/local/bin/common.sh /usr/local/bin/extract-vmlinux /usr/bin/unzboot /drivers \
       /mnt/shared-nvidia-driver-toolkit/
cp: cannot stat '/usr/bin/unzboot': No such file or directory
```

**Expected behavior**

DTK-mode build proceeds on amd64 (guard the `unzboot` entries by arch — copy list ~line 42 and usage ~lines 162/175 — or provide a no-op placeholder on amd64 at image build).

**Environment**

- `gpu-driver-container` source: image digest `sha256:7119e92a2a2f7b3286e764e5241294658dcfbe791a94e06d0807623be2e9c240` (580.178.04-rhel10.0); same code verified in `580.159.04-rhel10.0` and on `main`
- NVIDIA Driver Version: 580.178.04 (also 580.159.04)
- Host OS: CentOS Stream CoreOS 10 (SCOS)
- Kernel Version: 6.12.0-233.el10.x86_64
- Container Runtime: CRI-O 1.34
- CPU Architecture: x86_64
- GPU Model: NVIDIA L4
- Kubernetes Distro: OKD 4.22.0-okd-scos.6
- GPU Operator version: v26.3.3

lukiod:
Fixes #984.

`install.sh` only builds/installs `unzboot` on aarch64 (EPEL, zboot kernels). `ocp_dtk_entrypoint`'s shared-dir prep copied it unconditionally, so DTK mode crash-loops on amd64 before any build starts:

```
cp: cannot stat '/usr/bin/unzboot': No such file or directory
```

Guards both copy sites (the initial shared-dir prep in `nv-ctr-run-with-dtk`, and the later bin staging in `dtk-build-driver` which already has `$DRIVER_ARCH` computed) behind an aarch64 check, using the same `TARGETARCH`/`DRIVER_ARCH` normalization already used elsewhere in this file and in `install.sh`.

Verified by extracting both changed blocks and running them under bash with `TARGETARCH` set to each value: amd64 skips the copy cleanly with no missing-file error, arm64 with a real file present still copies it correctly.

### DIFF
--- a/rhel10/ocp_dtk_entrypoint
+++ b/rhel10/ocp_dtk_entrypoint
@@ -39,9 +39,15 @@ nv-ctr-run-with-dtk() {
            /usr/local/bin/nvidia-driver \
            /usr/local/bin/common.sh \
            /usr/local/bin/extract-vmlinux \
-           /usr/bin/unzboot \
            /drivers \
            "$DRIVER_TOOLKIT_SHARED_DIR/"
+
+        # unzboot is only built/installed on aarch64 (see install.sh); it doesn't
+        # exist on amd64 images, so only copy it there.
+        DRIVER_ARCH=${TARGETARCH/amd64/x86_64} && DRIVER_ARCH=${DRIVER_ARCH/arm64/aarch64}
+        if [[ "$DRIVER_ARCH" == "aarch64" ]]; then
+          cp /usr/bin/unzboot "$DRIVER_TOOLKIT_SHARED_DIR/"
+        fi
 
         if [[ -f "/usr/local/bin/vgpu-util" ]]; then
           cp /usr/local/bin/vgpu-util "$DRIVER_TOOLKIT_SHARED_DIR/"
@@ -172,8 +178,11 @@ dtk-build-driver() {
        "$DRIVER_TOOLKIT_SHARED_DIR/nvidia-driver" \
        "$DRIVER_TOOLKIT_SHARED_DIR/common.sh" \
        "$DRIVER_TOOLKIT_SHARED_DIR/extract-vmlinux" \
-       "$DRIVER_TOOLKIT_SHARED_DIR/unzboot" \
        "${DRIVER_TOOLKIT_SHARED_DIR}/bin"
+
+    if [[ "$DRIVER_ARCH" == "aarch64" ]]; then
+      cp -v "$DRIVER_TOOLKIT_SHARED_DIR/unzboot" "${DRIVER_TOOLKIT_SHARED_DIR}/bin"
+    fi
 
     if [[ -f "$DRIVER_TOOLKIT_SHARED_DIR/vgpu-util" ]]; then
       cp -v "$DRIVER_TOOLKIT_SHARED_DIR/vgpu-util" "$DRIVER_TOOLKIT_SHARED_DIR/bin"


myeolenv:
Thanks for the contribution. We will review this PR soon.

kvalliyurnatt (approved):

rhel10/ocp_dtk_entrypoint
```diff
@@ -39,10 +39,16 @@ nv-ctr-run-with-dtk() {
            /usr/local/bin/nvidia-driver \
            /usr/local/bin/common.sh \
            /usr/local/bin/extract-vmlinux \
-           /usr/bin/unzboot \
            /drivers \
            "$DRIVER_TOOLKIT_SHARED_DIR/"
 
+        # unzboot is only built/installed on aarch64 (see install.sh); it doesn't
+        # exist on amd64 images, so only copy it there.
+        DTK_PREP_ARCH=${TARGETARCH/amd64/x86_64} && DTK_PREP_ARCH=${DTK_PREP_ARCH/arm64/aarch64}
```
tariq1890: What is the purpose of this variable, `DTK_PREP_ARCH`?

lukiod: DTK_PREP_ARCH is TARGETARCH (amd64/arm64) translated to the uname style (x86_64/aarch64). install.sh already does this exact same translation for its own DRIVER_ARCH and gates the unzboot build on DRIVER_ARCH = aarch64 -> this just mirrors that check so the entrypoint only copies unzboot on the arch where install.sh actually built it.

tariq1890: I guess I am not convinced that we need to introduce a new variable here. Why can't we just use the `$DRIVER_ARCH` variable?

tariq1890: See [L150](https://github.com/NVIDIA/gpu-driver-container/blob/2957a8b39352c76bbdf0a80d6efa42616c464750/rhel10/ocp_dtk_entrypoint#L150)

lukiod: $DRIVER_ARCH is set inside dtk-build-driver(), not nv-ctr-run-with-dtk(). This script only runs one function per invocation ($command at the bottom), so they never share shell state. Reusing $DRIVER_ARCH here would hit set -eu's unbound-variable check and crash before it ever ran.

tariq1890: Got it. Thanks for the explanation

tariq1890: Can't we just use this variable name instead?
```suggestion
        DRIVER_ARCH=${TARGETARCH/amd64/x86_64} && DRIVER_ARCH=${DRIVER_ARCH/arm64/aarch64}
```

rhel10/ocp_dtk_entrypoint
```diff
@@ -39,10 +39,16 @@ nv-ctr-run-with-dtk() {
...
+        if [[ "$DTK_PREP_ARCH" == "aarch64" ]]; then
```
tariq1890:
```suggestion
        if [[ "$DRIVER_ARCH" == "aarch64" ]]; then
```

rhel10/ocp_dtk_entrypoint
tariq1890: As you mentioned in the earlier thread, the `DRIVER_ARCH` variable is not in scope in this method, so it will just be treated as a net new variable similar to `DTK_PREP_ARCH`, right?

lukiod: Right, no scope conflict since they never run in the same invocation. Renamed it, pushed.

tariq1890: Thanks!

tariq1890:
Thanks @lukiod !

These are the remaining action items:
- Squash the commit history.
- Commit passes the DCO check. See this [page](https://docs.pi-hole.net/guides/github/how-to-signoff/) to learn how to sign off commits. 
- Commit is signed with a GPG key. Refer to [this](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to learn more

PR #990, opened 2026-08-27
Build native archs and then use the native arch build to create the multi arch build

kvalliyurnatt:
Build standard driver images on native amd64 and arm64 runners instead of emulating arm64 with QEMU.
- Add architecture to the image build matrix.
- Run each build on its matching native runner.
- Publish architecture-specific images.
- Create the existing multi-architecture tags after both builds finish.
- Detect the host architecture for native-only local builds.
- Leave the precompiled image flow unchanged.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -24,14 +24,17 @@ on:
 
 jobs:
   image:
-    runs-on: linux-amd64-cpu4
+    runs-on: linux-${{ matrix.arch }}-cpu4
     strategy:
       matrix:
-        driver:
+        arch:
+          - amd64
+          - arm64
+        driver: &driver-versions
           - 580.178.04
           - 595.91.07
           - 610.57.04
-        dist:
+        dist: &distributions
           - ubuntu22.04
           - ubuntu24.04
           - ubuntu26.04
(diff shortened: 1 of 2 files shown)


.github/workflows/image.yaml
```diff
@@ -95,10 +91,43 @@ jobs:
       - name: Build image
         env:
           IMAGE_NAME: ghcr.io/${LOWERCASE_REPO_OWNER}/driver
-          VERSION: ${COMMIT_SHORT_SHA}
+          VERSION: ${COMMIT_SHORT_SHA}-${{ matrix.arch }}
         run: |
           DRIVER_VERSIONS=${{ matrix.driver }} make build-${{ matrix.dist }}-${{ matrix.driver }}
 
+  image-manifest:
+    if: github.actor != 'dependabot[bot]'
```
tariq1890, line 99: So If I understand correctly, we are just skipping the multi-arch manifest creation step, but proceeding with the actual image builds when the PR is created by dependabot?

kvalliyurnatt: right, this is because we don't push builds for dependabot changes today, so we won't have images to reference for the multi-arch manifest

.github/workflows/image.yaml
```diff
@@ -128,7 +149,7 @@ jobs:
           - 5.15
           - 6.8
         ispr:
-          - ${{github.event_name == 'pull_request'}}
+          - ${{ github.event_name == 'pull_request' || startsWith(github.ref_name, 'pull-request/') }}
```
kvalliyurnatt, line 194: This was not working as intended before because the workflow runs PR pipelines as pushes to pull-request/* branches, so all 32 precompiled builds were running as part of PRs, now this should be down to 8

PR #1004, opened 2026-09-09
Add support for driver 615.71.09

aryangorwade:
This PR adds support for driver 615.71.09.

NVIDIA does not use the `-dkms` model stream anymore for all driver versions >= 615. Only the `-open` stream remains. This PR contains the changes for a regular release plus special logic implementing this on rhel8/9 releases.


.nvidia-ci.yml
```diff
@@ -34,7 +34,7 @@ variables:
   # Define the public staging registry
   STAGING_REGISTRY: ghcr.io/nvidia
   STAGING_VERSION: "${CI_COMMIT_SHORT_SHA}"
-  PUBLISH_VERSIONS: 580.178.04 595.91.07 610.57.04
+  PUBLISH_VERSIONS: 580.178.04 595.91.07 610.57.04 615.71.09
```
rajathagasthya: This should be just:
```yaml
PUBLISH_VERSIONS: 615.71.09
```

aryangorwade: Right so PUBLISH_VERSIONS is only the versions being updated by this commit?

rajathagasthya: Essentially, yes. It contains only the version that will be published by the release pipeline.

tariq1890 (approved):

PR #1008, opened 2026-09-11
chore: remove 610.57.04 driver as it's EOL

tariq1890:


aryangorwade (approved):
Also updates go version from 1.26 to 1.27

aryangorwade:
Why is 610 EOL while older versions like 580 and 590 are not?

versions.mk
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 # DRIVER_VERSIONS contains latest version in all active datacenter branches
-DRIVER_VERSIONS ?= 580.178.04 595.91.07 610.57.04 615.71.09
+DRIVER_VERSIONS ?= 580.178.04 595.91.07 615.71.09
 
-GOLANG_VERSION := 1.26.6
+GOLANG_VERSION := 1.27.1
```
abrarshivani: Is this intentional?

abrarshivani (approved):

cdesiniotis:
> Why is 610 EOL while older versions like 580 and 590 are not?

See https://docs.nvidia.com/datacenter/tesla/drivers/latest/supported-drivers-and-cuda-toolkit-versions.html and https://docs.nvidia.com/datacenter/tesla/drivers/latest/driver-lifecycle.html. Different branches have different support lifecycles.

versions.mk
tariq1890: Yes, this was intentional. If you prefer that I do it in a separate commit, I can do that too

abrarshivani: Yes, I think that would be clean.

tariq1890: done

abrarshivani: Thanks

PR #1009, opened 2026-09-11
[precompiled]: remove unused pebble binary from Ubuntu 26.04 base image

abrarshivani:
The ubuntu:resolute-20260901 base image used for the ubuntu26.04 precompiled
driver container ships a pebble binary (/usr/bin/pebble, /var/lib/pebble)
that this image never uses. The container's entrypoint is nvidia-driver,
and pebble isn't invoked anywhere in the driver install or runtime scripts.

pebble is currently affected by CVE-2026-39821. Since it's dead weight in
this image, strip it out during the build rather than carrying the CVE
exposure forward.

Added a RUN step in ubuntu26.04/precompiled/Dockerfile that removes the
binary and its state directory right after the base package setup.

Verified locally that a build with this change no longer contains any
pebble files.


ubuntu26.04/precompiled/Dockerfile
```diff
@@ -40,6 +40,9 @@ RUN rm -f /etc/apt/sources.list.d/cuda* && \
 
 RUN usermod -o -u 0 -g 0 _apt
 
+# pebble ships in the base image but is unused here; ENTRYPOINT is nvidia-driver, not pebble
+RUN rm -rf /usr/bin/pebble /var/lib/pebble
```
tariq1890, line 44: If pebble is associated with a deb package, can we remove that instead of deleting every possible pebble file?

abrarshivani: It is not dpkg managed file.

ubuntu26.04/precompiled/Dockerfile
```diff
@@ -40,6 +40,11 @@ RUN rm -f /etc/apt/sources.list.d/cuda* && \
 
 RUN usermod -o -u 0 -g 0 _apt
 
+# pebble ships in the base image but is unused here; ENTRYPOINT is nvidia-driver, not pebble.
+# It isn't dpkg-managed (dpkg -S finds no owning package), so apt/dpkg can't remove it - rm is
+# the only option.
```
tariq1890:
```suggestion
# pebble ships in the base image but is unused here; ENTRYPOINT is nvidia-driver, not pebble.
```

tariq1890: These two lines are not needed. In my earlier comment, I meant to say "remove them"

abrarshivani: done

tariq1890 (approved):

PR #1017, opened 2026-09-16
Fix typos

antonkesy:
Fixes typos in comments and stdout texts which got copied around the whole repository


rajathagasthya (approved):
Thanks!

